PR #16, opened 2015-11-13
Fix AutoYaST schema in SLE 12 GA

imobach:
Fix AutoYaST schema. BTW, it improves the definition of some `yesno` elements.


schubi2:
LGTM

PR #20, opened 2016-02-05
Merge firewall configuration instead of overwriting it

imobach:
When using an AutoYaST profile, it merges firewall configuration instead of overwriting it. Second part of https://github.com/yast/yast-yast2/pull/436.


ancorgs:
LGTM

imobach:
Thanks!

PR #21, opened 2016-02-05
Merge into SP1

teclator:


src/clients/firewall_auto.rb
```diff
@@ -87,9 +87,7 @@ def main
         SuSEFirewall.SetStartService(SuSEFirewall.GetEnableService)
       # Import configuration
       elsif @func == "Import"
-        @ret = SuSEFirewall.Import(
-          Convert.convert(@param, :from => "map", :to => "map <string, any>")
-        )
+        @ret = SuSEFirewall.read_and_import(@param)
```
jreidinger: it returns void, so why store it?

jreidinger: and btw when I see usage, then maybe it is better to modify import to have by default read and import. E.G. bootloader is done in this way

imobach: We did it in that way to avoid changing the behavior of SuSEFirewall.Import which is public. I fear that someone could use it.

jreidinger: then calling import via client and via API will be different which is confusing. I think bug was original behavior so it should be changed.

imobach:
Superseeded by #23.

PR #22, opened 2016-02-05
Revert config merge

imobach:
Revert firewall configuration merge. As pointed by @jreidinger, we should modify SuSEFirewall.Import instead of using a new method. It was done in this way because we feared someone was using that method. But it can be confusing.


jreidinger:
LGTM

imobach:
Thanks!

imobach:
Thanks.

PR #24, opened 2016-05-24
FirewallD support

hwoarang:
Hi,

This completes the firewalld support in Yast2 (https://github.com/yast/yast-yast2/pull/471, https://github.com/yast/yast-yast2/pull/457) by adding support for this backend in the yast-firewall module. This patchset uses the upstream firewall-config UI instead of the one in the module. This is because I believe it's more beneficial to use what upstream provides instead of designing our own. However, since the SF2 and FirewallD APIs are (almost) compatible, it's should be possible to use the existing one without too much effort. However, using the upstream one means that no ncurses support is provided. Those who prefer to configure Yast2 from their terminal, it's possible to use the Yast2 command line interface, or better yet, use the upstream ```firewall-cmd``` tool.

### DIFF
--- a/package/yast2-firewall.spec
+++ b/package/yast2-firewall.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        3.1.4
+Version:        3.1.5
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
@@ -30,8 +30,8 @@ BuildRequires:  yast2-devtools >= 3.1.10
 # IP::CheckNetwork
 BuildRequires:	yast2 >= 2.23.25
 
-# IP::CheckNetwork
-Requires:	yast2 >= 2.23.25
+# FirewallD backend
+Requires:	yast2 >= 3.1.191
 
 # ButtonBox widget
 Conflicts:	yast2-ycp-ui-bindings < 2.17.3
--- a/src/clients/firewall.rb
+++ b/src/clients/firewall.rb
@@ -30,6 +30,9 @@
 # $Id$
 #
 # File includes helps for yast2-firewall dialogs.
+#
+require "network/susefirewalld"
+
 module Yast
   class FirewallClient < Client
     def main
@@ -57,11 +60,27 @@ module Yast
         SuSEFirewallCMDLine.Run 
         # GUI or TextUI
       else
-        # installation has other sequence
-        if Mode.installation
-          @ret = FirewallInstallationSequence()
+        # If FirewallD then use it's UI
+        if SuSEFirewall.is_a?(Yast::SuSEFirewalldClass)
+          # We can't do ncurces. Lets see if the firewalld-config
+          # is installed
+          Yast.import "PackageSystem"
+          Yast.import "UI"
+          if UI.TextMode()
+            Yast::Popup.Error(_("Your display can't support the 'firewall-config' UI.\n") +
+                               _("Either use the Yast2 command line or the 'firewall-cmd' utility.") )
+            return false
+          end
+          if PackageSystem.CheckAndInstallPackages(["firewall-config"])
+            @ret = SCR.Execute(Yast::Path.new(".target.bash"), "/usr/bin/firewall-config")
+          end
         else
-          @ret = FirewallSequence()
+          # installation has other sequence
+          if Mode.installation
+            @ret = FirewallInstallationSequence()
+          else
+            @ret = FirewallSequence()
+          end
         end
       end
 
--- a/src/modules/SuSEFirewallCMDLine.rb
+++ b/src/modules/SuSEFirewallCMDLine.rb
@@ -1645,7 +1645,7 @@ module Yast
       return unless firewalld?
 
       # Actions not supported by FirewallD
-      firewalld_disabled = ["broadcast"]
+      firewalld_disabled = ["broadcast", "masqredirect"]
 
       firewalld_disabled.each do |opt|
         @cmdline["actions"].delete(opt)
--- a/src/modules/SuSEFirewallCMDLine.rb
+++ b/src/modules/SuSEFirewallCMDLine.rb
@@ -377,6 +377,9 @@ module Yast
           "disable"      => []
         }
       }
+
+      ConfigureFirewalld()
+
     end
 
     # Returns list of strings made from the comma-separated string got as param.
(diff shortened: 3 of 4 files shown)


src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1534,30 +1546,53 @@ def FWCMDMasqRedirect(options)
     # @return [Boolean] whether write call is needed
     def FWCMDMasquerade(options)
       options = deep_copy(options)
+      zone = nil
+      if firewalld?
+        if Ops.get(options, "zone") != nil
```
jreidinger: can be written as 
```
if options["zone"]
```

src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1534,30 +1546,53 @@ def FWCMDMasqRedirect(options)
...
         CommandLine.Print("")
         # TRANSLATORS: CommandLine header
         CommandLine.Print(
           String.UnderlinedHeader(_("Masquerading Settings:"), 0)
         )
         CommandLine.Print("")
+
+       # TRANSLATORS: CommandLine informative text, either "everywhere" or
+       # "in the %1 zone" where %1 is zone name.
+       zone_msg = zone == nil ? "everywhere" :
```
jreidinger: everywhere is not marked for translation with `_()` method

src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1599,6 +1634,40 @@ def Run
       nil
     end
 
+  private
+    # Returns true if FirewallD is the running backend
+    def firewalld?
+      return true if SuSEFirewall.is_a?(Yast::SuSEFirewalldClass)
+
+      false
+    end
```
jreidinger: can be one liner with `SuSEFirewall.is_a?(Yast::SuSEFirewalldClass)`

src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1599,6 +1634,40 @@ def Run
...
+
+    def ConfigureFirewalld
+      return if !firewalld?
```
jreidinger: in ruby is convention to use unless for single statement trailing condition, it looks like `return unless firewalld?`

package/yast2-firewall.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        3.1.4
+Version:        3.1.5
```
jreidinger: below in spec file you have to increase version for yast2 dependecy to ensure you have yast2 with firewalld support

src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1599,6 +1634,40 @@ def Run
       nil
     end
 
+  private
+    # Returns true if FirewallD is the running backend
+    def firewalld?
+      return true if SuSEFirewall.is_a?(Yast::SuSEFirewalldClass)
```
jreidinger: this depends on hidden dependecy of SuSEFirewall on Yast::SuSEFirewalldClass, I prefer explicit requirement of Yast::SuSEFirewalldClass class for this comparison

hwoarang: @jreidinger I don't understand that. Could you explain what you mean? SuSEFirewall is an object which is instantiated in SuSEFirewall.rb. It's type can either be SuSEFirewalldClass or SuSEFirewall2Class. Both of these classes are being pulled in in SuSEFirewall.rb. So I am not sure what's wrong with that...

jreidinger: well, problem is that it is not obvious that SuSEFirewalldClass is pulled by SuSEFirewall.rb, so I will prefer to have here require of SuSEFirewalld file ( ruby ensure it is not loaded twice, no worry ). It is just if we decide in future that SuSEFirewall do not require SuSEFirewalldClass it won't break here.

hwoarang: aha I see. OK I will fix that thanks!

src/clients/firewall.rb
```diff
@@ -57,11 +57,34 @@ def main
         SuSEFirewallCMDLine.Run 
         # GUI or TextUI
       else
-        # installation has other sequence
-        if Mode.installation
-          @ret = FirewallInstallationSequence()
+        # If FirewallD then use it's UI
+        if SuSEFirewall.is_a?(Yast::SuSEFirewalldClass)
```
jreidinger: ah, now I see that here is also used suseFirewalldClass, so also here should be require at top

src/clients/firewall.rb
```diff
@@ -57,11 +57,34 @@ def main
...
+          if UI.TextMode()
+            UI.OpenDialog(
+              VBox(
+                Label(_("Your display can't support the 'firewall-config' UI.")),
+                Label(_("Either use the Yast2 command line or the 'firewall-cmd' utility.")),
+                PushButton( Opt(:default), "OK" ),
+              )
+            )
+            UI::UserInput();
+            UI::CloseDialog();
```
jreidinger: Not needed, you can use already existing popup class - https://github.com/yast/yast-yast2/blob/8eb9214265fdf8afcae1531c5a0309799028d79d/library/general/src/modules/Popup.rb#L1439

so it can look like
```
if UI.TextMode()
  Yast::Popup.Notify(_("Your display can't support the 'firewall-config' UI.") + "<br>" + _("Either use the Yast2 command line or the 'firewall-cmd' utility.") )
  return false
end
```

or maybe warning or error can be option for you.

src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1534,30 +1547,53 @@ def FWCMDMasqRedirect(options)
     # @return [Boolean] whether write call is needed
     def FWCMDMasquerade(options)
       options = deep_copy(options)
+      zone = nil
+      if firewalld?
+        if options["zone"] != nil
```
jreidinger: still `!= nil` is not needed, nil is false value, so it can be simple `if options["zone"]`

src/modules/SuSEFirewallCMDLine.rb
```diff
@@ -1534,30 +1547,53 @@ def FWCMDMasqRedirect(options)
...
+          zone = Builtins.tolower(Ops.get_string(options, "zone"))
```
jreidinger: here you can also replace get_string with ruby access like `options["zone"]`

jreidinger: and also tolower is not needed now in ruby, co whole line can look like
```
zone = options["zone"].downcase
```

jreidinger:
Thanks for changes. Tested manually and it works for SuSEFirewall2, so lets merge it.

PR #27, opened 2017-04-10
bsc#962713: Inform the user if start/stop/restart fails

gilsonsouza:


src/include/firewall/uifunctions.rb
```diff
@@ -135,21 +144,33 @@ def SaveAndRestart
...
     def StopNow
       UI.OpenDialog(Label(_("Stopping firewall...")))
       SuSEFirewall.SetStartService(false)
-      ret = SuSEFirewall.StopServices
+      success = SuSEFirewall.StopServices
+      Popup.Error(
+        _(
+          "The firewall could not be stopped.\n" +
+          "Please verify your system and try again."
+        )
```
jreidinger: well, it looks a bit duplicite for maybe have one method for it? like

```
def report_problem(action)
action_str = case action
when :stop
  # TRANSLATORS: action which failed, used later
  _("stopped")
when :start
  # TRANSLATORS: action which failed, used later
  _("started")
else
  raise "invalid action #{action}"
end
# TRANSLATORS: %s is action that failed
msg = _(
"The firewall could not be %s.\n" +
"Please verify your system and try again."
) % action_str
Popup.Error(msg)
end

# and usage
report_problem(:start)
```

lslezak: On the other hand the original version is more generic, imagine a language where words "start" and "stop" use a different grammar rules and a simple `%s` placeholder might not be enough.

The question is whether we want to be rather more generic or make the work for translators easier...

jreidinger: yeah, but consider that also enable failed, disable failed, restart failed, reload failed or something else. I worry it can be a bit mess later.

mvidner: Composing sentences is very tricky if you have to do it across languages, as illustrated with a quick IRC experiment:
```
<a> b:  # TRANSLATORS: action which failed, used later
<a> b:  _("stopped")
<a> b: now, without looking at the PR, write the correct Czech translation here
<b> a: "zastavil"?
<b> or "zastaveni"
<a> b: wrong, thank you for proving my point :)
<a> the correct string is Zastavení
<b> so second one is correct :)
<a> no, you got the capitalization wrong
```

If we want to factor out parts of translatable strings, I would do it at sentence boundaries to minimize problems.

lslezak: Yes, having separate `_("Please verify your system and try again.")` will help with the duplication.

PR #28, opened 2017-04-11
Backport master into Sle 12 sp1

gilsonsouza:


package/yast2-firewall.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        3.1.4
+Version:        3.1.5
```
jreidinger: this is wrong, 3.1.5 already existed in SP2. It have to be 3.1.4.1

PR #32, opened 2017-06-14
Do not crash with internal error when adding from empty "Services to allow" list

mchf:
Bnc1044045


kobliha:
LGTM, thx

mchf:
Thanks for the review

PR #36, opened 2017-11-08
Fixed method call and added tests for installation client.

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1067320

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -3,7 +3,6 @@ Thu Nov  9 09:16:14 UTC 2017 - knut.anderssen@suse.com
 
 - ProposalSettings: fixed method name when vnc is enable by Linuxrc
   (bsc#1067320)
-- Added tests for installation client.
 - 4.0.2
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


package/yast2-firewall.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Nov  9 09:16:14 UTC 2017 - knut.anderssen@suse.com
+
+- ProposalSettings: fixed method name when vnc is enable by Linuxrc
+  (bsc#1067320)
+- Added tests for installation client.
```
imobachgs: I would not mention it in the changes file (tests are not shipped with the package).

test/clients/installation_finish_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+describe Y2Firewall::Clients::InstallationFinish do
+  describe "#title" do
+    it "returns translated string" do
+      expect(subject.title).to be_a(::String)
+    end
+  end
+
+  describe "#modes" do
+    it "runs in installation" do
```
imobachgs: np: You could say "it runs in installation and autoinstallation" (or something similar) and write only one test:

```ruby
excpect(subject.modes).to eq([:installation, :autoinst])
```

We can leave it for later.

PR #37, opened 2017-11-09
Attr error (bsc#1067320)

teclator:
- Hope to fix the last issue with the Firewall & SSH proposal, covered now by tests.

### DIFF
--- a/src/lib/y2firewall/proposal_settings.rb
+++ b/src/lib/y2firewall/proposal_settings.rb
@@ -75,7 +75,7 @@ module Y2Firewall
 
     def load_feature(feature, to, source: global_section)
       value = Yast::Ops.get(source, feature.to_s)
-      send("#{to}=", value) unless value.nil?
+      public_send("#{to}=", value) unless value.nil?
     end
 
     def global_section
--- a/src/lib/y2firewall/proposal_settings.rb
+++ b/src/lib/y2firewall/proposal_settings.rb
@@ -68,14 +68,14 @@ module Y2Firewall
 
     def close_vnc!
       log.info "Close VNC port"
-      self.close_vnc = true
+      self.open_vnc = false
     end
 
   private
 
     def load_feature(feature, to, source: global_section)
       value = Yast::Ops.get(source, feature.to_s)
-      self.send("#{to}=", value) unless value.nil?
+      send("#{to}=", value) unless value.nil?
     end
 
     def global_section
--- a/test/proposal_settings_test.rb
+++ b/test/proposal_settings_test.rb
@@ -160,8 +160,7 @@ describe Y2Firewall::ProposalSettings do
 
   describe "#open_ssh!" do
     it "sets the ssh port to be opened" do
-      subject.close_ssh!
-      expect(subject.open_ssh).to be(false)
+      subject.open_ssh = false
       subject.open_ssh!
       expect(subject.open_ssh).to be(true)
     end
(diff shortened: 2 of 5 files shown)


test/proposal_settings_test.rb
```diff
@@ -0,0 +1,195 @@
...
+    it "sets the ssh port to be closed" do
+      subject.open_ssh!
+      expect(subject.open_ssh).to be(true)
+      subject.close_ssh!
+      expect(subject.open_ssh).to be(false)
+    end
+  end
+
+  describe "#open_vnc!" do
+    it "sets the vnc port to be opened" do
```
mchf: most probably copy - paste issue here

test/proposal_settings_test.rb
```diff
@@ -0,0 +1,195 @@
...
+      subject.disable_sshd!
+    end
+  end
+
+  describe "#open_ssh!" do
+    it "sets the ssh port to be opened" do
+      subject.close_ssh!
+      expect(subject.open_ssh).to be(false)
+      subject.open_ssh!
+      expect(subject.open_ssh).to be(true)
```
mchf: hugh. I don't understand the code well -> most probably will need to go deeply. 

I'm guessing here, but is ```open_ssh``` just a status checker? Shouldn't it be ```opened_ssh?``` instead

argh it is attribute accessor. So, 1) shouldn't be attr_reader enough for ```open_ssh```? Is breaking the encapsulation really needed? 2) this workflow ```close_ssh!``` -> ```open_ssh``` --[port remains closed]--> ```open_ssh!``` --[port finally opened] looks a little bit weird at first view.

src/lib/y2firewall/proposal_settings.rb
```diff
@@ -68,14 +68,14 @@ def open_vnc!
...
-      self.close_vnc = true
+      self.open_vnc = false
     end
 
   private
 
     def load_feature(feature, to, source: global_section)
       value = Yast::Ops.get(source, feature.to_s)
-      self.send("#{to}=", value) unless value.nil?
+      send("#{to}=", value) unless value.nil?
```
jreidinger: public send is not enough? readers are public.

test/proposal_settings_test.rb
```diff
@@ -0,0 +1,196 @@
...
+      subject.disable_firewall!
+    end
+  end
+
+  describe "#enable_sshd!" do
+    it "sets sshd service to be enabled" do
+      allow(Yast::PackagesProposal).to receive("AddResolvables")
+        .with("firewall", :package, ["openssh"])
+
+      subject.enable_sshd!
```
jreidinger: when seeing this, what is reason to have these methods with "!" ?

teclator: well if it changes the internal state of the object then the convention is to use "!" isn't it?

test/proposal_settings_test.rb
teclator: I could to initializing the attributes, but also considered that it abstract from initialization and you at the end test that both methods are the opposite but as you prefer

test/proposal_settings_test.rb
jreidinger: well, see discussion at https://stackoverflow.com/questions/7179016/what-is-the-purpose-of-and-at-the-end-of-method-names
If it is surprising or dangerous operation it should have "!". For me it indicates like it it try to enable it now and can raise if failed ( but maybe just my POV, we can ask someone else how he see it ). @mvidner @lslezak or @imobachgs ?

teclator: I'm fine with the discussion. As this implementation was mainly for the POC, what if we maintain the discussion and maybe change the design but merge this PR that is now covering with unit tests the code and then apply the discussion results in next PR.

Are you fine with that?

teclator: And well about surprising or dangerous, it is not always the case: usually the exclamation mark (!)  means the object is modified like with `downcase!` `sort!` and so on while `downcase` `sort`don't modify the object.

PR #38, opened 2017-11-15
Services are not touched if firewalld is not enabled (bsc#1067915)

teclator:
- Fixed https://bugzilla.suse.com/show_bug.cgi?id=1067915
- Also fixed a bad method call in SuSEFirewallCMDLine
- Improved test coverage.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,39 @@
+# use the shared YaST defaults
+inherit_from:
+  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
+
+# Redundant returns add legibility for developers used to other languages
+Style/RedundantReturn:
+  Enabled: false
+
+# Don't enforce any particular name for block params
+SingleLineBlockParams:
+  Enabled: false
+
+# The general YaST agreement is 100 characters, so it fits into standard
+# resolutions and Github's pull request view. But let's open the threshold a
+# little bit
+Metrics/LineLength:
+  Max: 105
+
+# Enforce if/unless at the end only for really short lines
+Style/IfUnlessModifier:
+  MaxLineLength: 60
+
+# some storage API have size method, but without empty? method
+# for details see https://github.com/yast/yast-storage-ng/pull/83
+Style/ZeroLengthPredicate:
+  Enabled: false
+
+# the ".freeze" attribute for the constants is not nice
+Style/MutableConstant:
+  Enabled: false
+
+Metrics/CyclomaticComplexity:
+  Max: 10
+
+AllCops:
+  Exclude:
+    - "src/modules/*.rb"
+    - "src/clients/*.rb"
+    - "src/include/*/*.rb"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,38 +2,11 @@
 inherit_from:
   /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
 
-# Redundant returns add legibility for developers used to other languages
-Style/RedundantReturn:
-  Enabled: false
-
-# Don't enforce any particular name for block params
-SingleLineBlockParams:
-  Enabled: false
-
-# The general YaST agreement is 100 characters, so it fits into standard
-# resolutions and Github's pull request view. But let's open the threshold a
-# little bit
-Metrics/LineLength:
-  Max: 105
-
-# Enforce if/unless at the end only for really short lines
-Style/IfUnlessModifier:
-  MaxLineLength: 60
-
-# some storage API have size method, but without empty? method
-# for details see https://github.com/yast/yast-storage-ng/pull/83
-Style/ZeroLengthPredicate:
-  Enabled: false
-
-# the ".freeze" attribute for the constants is not nice
-Style/MutableConstant:
-  Enabled: false
-
 Metrics/CyclomaticComplexity:
   Max: 10
 
 AllCops:
   Exclude:
     - "src/modules/*.rb"
-    - "src/clients/*.rb"
+    - "src/clients/firewall_auto.rb"
     - "src/include/*/*.rb"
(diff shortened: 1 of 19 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,39 @@
+# use the shared YaST defaults
+inherit_from:
+  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
+
+# Redundant returns add legibility for developers used to other languages
+Style/RedundantReturn:
+  Enabled: false
```
jreidinger: btw reason sounds a bit funny. Does we add to C++ also style that makes C++ more familiar for ruby developers? :)

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+# resolutions and Github's pull request view. But let's open the threshold a
+# little bit
+Metrics/LineLength:
+  Max: 105
+
+# Enforce if/unless at the end only for really short lines
+Style/IfUnlessModifier:
+  MaxLineLength: 60
+
+# some storage API have size method, but without empty? method
```
jreidinger: this is not storage :)

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+# for details see https://github.com/yast/yast-storage-ng/pull/83
+Style/ZeroLengthPredicate:
+  Enabled: false
+
+# the ".freeze" attribute for the constants is not nice
```
jreidinger: hmm, we have this discussion also in storage and in fact, freeze is what make constant real constant

.rubocop.yml
teclator: Hehe, nothing like use the current config of some new module and not modify comments 👎

.rubocop.yml
teclator: Just copied the rubocop config from what I considered the newest module and only added the exclude of old yast-firewall modules because of all the complains so. So will check the other options

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+Style/MutableConstant:
+  Enabled: false
+
+Metrics/CyclomaticComplexity:
+  Max: 10
+
+AllCops:
```
jreidinger: this is a bit strong, but ok for me.

teclator: Yep, it is, I will remove it gradually, by now I only wanted to cover the new code.

src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -45,15 +52,20 @@ def modes
 
       def write
         Service.Enable("sshd") if @settings.enable_sshd
-        @firewalld.enable! if @settings.enable_firewall
+
+        return true unless @settings.enable_firewall
```
jreidinger: so if firewall is not enabled, we did not disable it?

test/clients/firewall_test.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+
+        it "returns false" do
+          allow(Yast::Popup).to receive("Error")
+
+          expect(subject.run).to eq(false)
+        end
+      end
+
+      context "and the firewall-config package is installed" do
```
jreidinger: missing case when it is not installed.

jreidinger: in fact it is two cases as user is asked to install it and he can install or cancel.

test/clients/proposal_test.rb
```diff
@@ -0,0 +1,99 @@
...
+# find current contact information at www.suse.com.
+
+require_relative "../test_helper.rb"
+require "y2firewall/clients/proposal"
+
+describe Y2Firewall::Clients::Proposal do
+  let(:client) { described_class.new }
+
+  describe "#initialize" do
+    it "instantiates a new proposal settings" do
```
jreidinger: well, I think it is not needed to test that it is instantiated, as it is common ruby functionality.

src/lib/y2firewall/clients/installation_finish.rb
teclator: Well, this client is the only one that enables the service, make sense to disable it during installation?

test/clients/firewall_test.rb
teclator: true

src/lib/y2firewall/clients/installation_finish.rb
jreidinger: well, it depends if user click on disable firewall and product have it enabled by default, I expect it will be disabled.

test/clients/proposal_test.rb
teclator: well, it is part of the code and it is public, it is a simple test but even so..

teclator: BTW, this is really a test for lib/y2firewall/clients/proposal_test.rb so will move all the lib tests to a folder that really match its path

src/lib/y2firewall/proposal_settings.rb
```diff
@@ -1,11 +1,44 @@
...
 
 module Y2Firewall
+  # Class that stores the proposal settings for firewalld during installation.
   class ProposalSettings
     include Yast::Logger
     include Yast::I18n
-    attr_accessor :enable_firewall, :enable_sshd, :open_ssh, :open_vnc
 
+    # [Boolean] Whether the firewalld service will be enable
+    attr_accessor :enable_firewall
```
jreidinger: as you are having more rich enable/disable variants, does it make sense to still have public `enable_firewall=`?

teclator: Probably not make sense all, is preferable to not expose it, but currently it is used in `load_feature` although I could just modify the `public_send` by `an instance_variable_set` in bang methods and also in some tests.

teclator: Will try to replace in next iteration over code.

PR #42, opened 2018-01-09
AutoYaST: Support to set default zone and export configuration

teclator:
- [Trello Card](https://trello.com/c/jL8fhEKw/1091-5-firewall-firewalld-is-the-new-susefirewall-autoyast-support)

It depends on yast/yast-yast2#665

### DIFF
--- a/src/lib/y2firewall/importer_strategies/suse_firewall.rb
+++ b/src/lib/y2firewall/importer_strategies/suse_firewall.rb
@@ -103,10 +103,12 @@ module Y2Firewall
       end
 
       # Obtain the interfaces for the given SuSEFIrewall2 zone name from the
-      # profile.
+      # profile. It removes 'any' from the list of interfaces as it is an
+      # especial case for SuSEFirewall2.
       #
       # @param zone_name [String]
-      # @return [Array<String>, nil]
+      # @return [Array<String>, nil] return the list of interfaces without
+      # especial wildcards like 'any' or nil in case the key is not defined
       def interfaces(zone_name)
         interfaces = profile["FW_DEV_#{zone_name}"]
         interfaces ? interfaces.split(" ").reject { |i| i == "any" } : nil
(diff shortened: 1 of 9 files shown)


src/lib/y2firewall/importer_strategies/suse_firewall.rb
```diff
@@ -101,18 +106,39 @@ def services(zone_name)
       # profile.
       #
       # @param zone_name [String]
+      # @return [Array<String>, nil]
       def interfaces(zone_name)
         interfaces = profile["FW_DEV_#{zone_name}"]
-        interfaces ? interfaces.split(" ") : nil
+        interfaces ? interfaces.split(" ").reject { |i| i == "any" } : nil
```
imobachgs: Why not return always an `[]` or `nil` when no interfaces are found? I mean, it will return different values if interfaces are set to `any` (they will be filtered out anyway) and if no interfaces are found. Does it really matter?

It is not a big problem, as it is part of the private API, but still...

test/lib/y2firewall/importer_strategies/suse_firewall.rb
```diff
@@ -73,14 +73,18 @@
...
+      end
+
       context "and masquerade is disabled" do
         let(:masquerade) { "no" }
 
         it "configures the EXT zone as the public" do
           public_zone = firewalld.find_zone("public")
 
           expect(public_zone.interfaces).to eq(["eth0"])
-          expect(public_zone.services).to eq(["dhcp", "ssh", "samba", "vnc-server"])
```
imobachgs: Any reason to not check services?

teclator: I removed it because I checked the behavior in the next test, so not a strong one.

src/lib/y2firewall/importer_strategies/suse_firewall.rb
teclator: Well, nil is returned when the key is not defined, in other case an array is returned.

Just needed to remove special cases.

teclator:
@imobachgs I have added also export support

src/lib/y2firewall/importer_strategies/suse_firewall.rb
imobachgs: Then you should document that behaviour (or unify it). What about documenting this behaviour in the `@return` tag?

PR #45, opened 2018-01-31
Try to import again when writing if it failed prefiously

teclator:
- [Trello Card](https://trello.com/c/jL8fhEKw/1091-5-firewall-firewalld-is-the-new-susefirewall-autoyast-support)

1. During the second stage, an Import takes place before the Write. When the list of packages needed for configuring the firewall are not present, them are installed after the Import and we need them before it.
  This PR tries to import again when writing if the import was not completed previously.

2. The firewalld service is enabled or disabled depending on the profile
3. Fixed the requirement of packages to be installed (wrong implementation) see yast/yast-yast2#681

It depends on yast/yast-yast2#684

### DIFF
--- a/src/lib/y2firewall/clients/auto.rb
+++ b/src/lib/y2firewall/clients/auto.rb
@@ -40,8 +40,10 @@ module Y2Firewall
         # successfully or not
         attr_accessor :imported
         # @return [Boolean] whether the firewalld service has to be enabled
+        # after writing the configuration
         attr_accessor :enable
         # @return [Boolean] whether the firewalld service has to be started
+        # after writing the configuration
         attr_accessor :start
         # @return [Hash]
         attr_accessor :profile
@@ -117,7 +119,7 @@ module Y2Firewall
       end
 
       # A map with the packages that needs to be installed or removed for
-      # configuring properly firewalld
+      # configuring firewalld properly
       #
       # @return packages [Hash{String => Array<String>} ] of packages to be
       # installed or removed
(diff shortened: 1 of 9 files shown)


src/lib/y2firewall/clients/auto.rb
```diff
@@ -30,28 +31,55 @@ module Clients
     # goes through the configuration and return the setting.
     # Does not do any changes to the configuration.
     class Auto < ::Installation::AutoClient
+      include Yast::Logger
       class << self
         attr_accessor :changed
+        attr_accessor :imported
```
imobachgs: Please, document these accesors.

src/lib/y2firewall/clients/auto.rb
```diff
@@ -62,16 +90,30 @@ def change
         :next
       end
 
+      # Write the imported configuration to firewalld. If for some reason the
+      # configuration was not imported from the profile, it tries to import
+      # it again.
       def write
+        return false if !firewalld.installed?
+        import(self.class.profile) unless self.class.imported
```
imobachgs: What about defining an `imported?` method? Just to keep consistence with `installed?`.

src/lib/y2firewall/clients/auto.rb
```diff
@@ -62,16 +90,30 @@ def change
...
       def read
-        firewalld.read if firewalld.installed?
+        return false if !firewalld.installed?
+        firewalld.read
       end
 
+      # A map with the packages that needs to be installed or removed for
+      # configuring properly firewalld
+      #
+      # @return packages [Hash] of packages to be installed or removed
```
imobachgs: JFYI, you could do `Hash{String => Array<String>}` if I am not wrong (see https://www.yardoc.org/types.html).

src/lib/y2firewall/clients/auto.rb
```diff
@@ -84,13 +126,23 @@ def modified?
 
     private
 
+      # Depending on the profile it activate or not the firewalld service
```
imobachgs: *it activates*. On the other hand, you could say something like *it activates or deactivates* (because "not activate" != "deactivate").

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -49,22 +51,48 @@
...
+        expect(importer).to receive(:import).with(arguments)
+
+        subject.import(arguments)
+      end
+
+      it "returns true if import success" do
+        expect(subject.import(arguments)).to eq(true)
+      end
+
+      it "marks the importation as done or completed" do
```
imobachgs: Why not just *done* or * completed* (I mean only one of them)?

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -86,4 +114,39 @@
...
+      { "FW_MASQUERADE" => "yes", "enable_firewall" => false, "start_firewall" => false }
+    end
+
+    it "returns false if firewalld is not installed" do
+      expect(firewalld).to receive(:installed?).and_return(false)
+
+      expect(subject.write).to eq(false)
+    end
+
+    it "tries to import again the profile if was not imported" do
```
imobachgs: *if **it** was*

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -86,4 +114,39 @@
...
+      allow(subject.class).to receive(:profile).and_return(arguments)
+      expect(subject).to receive(:import).with(arguments).and_return(false)
+
+      subject.write
+    end
+
+    it "writes the configuration imported" do
```
imobachgs: I would say *imported configuration* (the adjetive comes first).

src/lib/y2firewall/clients/auto.rb
```diff
@@ -143,6 +153,33 @@ def firewalld
...
+      def enable
+        self.class.enable = true
+      end
+
+      # Whether the firewalld service has to be enable or not
+      def enable?
+        !!self.class.enable
+      end
+
+      def start
```
imobachgs: Please, document.

PR #46, opened 2018-02-05
Show error message when using ncurses or command line

gilsonsouza:


src/lib/y2firewall/clients/firewall.rb
```diff
@@ -36,20 +34,23 @@ def initialize
...
+        " firewall.\nPlease, either use \"firewall-config\" to configure your firewall" \
+        " via a user interface\nor \"firewall-cmd\" for the command line.".freeze
+
       def run
         log_and_return do
           if !Yast::WFM.Args.empty?
-            Yast.import "SuSEFirewallCMDLine"
-            Yast::SuSEFirewallCMDLine.Run
-            nil
+            puts _(NOT_SUPPORTED)
```
jreidinger: what about `$stderr.puts` ?

src/lib/y2firewall/clients/firewall.rb
```diff
@@ -36,20 +34,23 @@ def initialize
         Yast.import "UI"
         Yast.import "Popup"
         Yast.import "PackageSystem"
+        Yast.import "CommandLine"
 
         textdomain "firewall"
       end
 
+      NOT_SUPPORTED = "YaST currently does not have a module for configuring" \
```
jreidinger: this won;t work with translation. You have to use `N_` to mark it for markdown. Also some comment for translator would be nice ( like not translation stuff in quotes )

src/lib/y2firewall/clients/firewall.rb
```diff
@@ -36,20 +34,23 @@ def initialize
         Yast.import "UI"
         Yast.import "Popup"
         Yast.import "PackageSystem"
+        Yast.import "CommandLine"
```
jreidinger: you add import, but does not use it ( at least not in diff )

src/lib/y2firewall/clients/firewall.rb
```diff
@@ -40,16 +39,20 @@ def initialize
         textdomain "firewall"
       end
 
+      # TRANSLATORS: firewall-config and firewall-cmd are the name of software utilities,
```
jreidinger: are the name*s*

src/lib/y2firewall/clients/firewall.rb
```diff
@@ -40,16 +39,20 @@ def initialize
...
+      # so they should not be translated.
+      NOT_SUPPORTED = N_("YaST currently does not have a module for configuring" \
```
jreidinger: In general I suggest here to remove `\n` as for translators it is hard to place it properly to have reasonable line lenght and instead use where make sense later this method https://github.com/yast/yast-yast2/blob/5a2d13714e18c295f7e7973a23c4a872470c3bf4/library/general/src/lib/ui/text_helpers.rb#L33 on result of translation.

PR #48, opened 2018-02-07
Extended AY's summary

mchf:
1) prepare a layout [done]
   - high level overview + per zone details
2) refactoring [done]
  - used Yast::HTML to have formatting consistent with other yast modules
  - use higher level firewalld api (with caching) instead of direct calls to firewalld
3) enable / fix / improve testsuite [done]

firewalld_summary

### DIFF
--- a/src/lib/y2firewall/clients/auto.rb
+++ b/src/lib/y2firewall/clients/auto.rb
@@ -60,7 +60,20 @@ module Y2Firewall
       def summary
         return "" if !firewalld.installed?
 
-        firewalld.api.list_all_zones.join("\n")
+        default_zone = firewalld.api.default_zone
+        zones = firewalld.api.zones
+
+        summary = "<p>"
+        summary << "<b>Default ZONE:</b> #{default_zone}"
+        summary << "<b>Defined zones:</b>"
+        summary << "<ul>" + zones.map { |z| "<li />#{z}" }.join("\n") + "</ul>"
+        summary << "</p>"
+
+        zones.each do |zone|
+          summary << zone_summary(zone)
+        end
+
+        summary
       end
 
       # Import the firewall configuration
@@ -197,6 +210,38 @@ module Y2Firewall
       # @return [Boolean] true if has been imported; false otherwise
       def imported?
         !!self.class.imported
+      end
+
+
+      FIREWALLD_ATTRS = [
+        "interfaces".freeze,
+        "services".freeze,
+        "ports".freeze,
+        "protocols".freeze
+      ]
+      FIREWALLD_ATTRS.each do |attr|
+        define_method("#{attr}_summary") do |zone|
+          status = firewalld.api.send("list_#{attr}", zone)
+          return "" if status.empty?
+
+          status.map! { |s| "<li />#{s}" }
+          "<li /><b>#{attr.capitalize}:</b> " + "<ul>" + status.join("\n") + "</ul>"
+        end
+      end
+
+      # Creates a summary for the given zone
+      #
+      # @param [String] zone name of zone
+      # @return [String] HTML formated zone description
+      def zone_summary(zone)
+        return "" if zone.nil? || zone.empty?
+
+        desc = ""
+        FIREWALLD_ATTRS.each { |attr| desc << send("#{attr}_summary", zone) }
+        return "" if desc.empty?
+
+        summary = "<h3>#{zone}</h3>"
+        summary << "<ul>#{desc}</ul>"
       end
     end
   end
--- a/src/lib/y2firewall/clients/auto.rb
+++ b/src/lib/y2firewall/clients/auto.rb
@@ -63,16 +63,11 @@ module Y2Firewall
       def summary
         return "" if !firewalld.installed?
 
-        default_zone = firewalld.api.default_zone
-        zones = firewalld.api.zones
-
         # general overview
-        summary = Yast::HTML.Bold("Default ZONE:") + " #{default_zone}"
-        summary << Yast::HTML.Bold("Defined zones:")
-        summary << Yast::HTML.List(zones)
-        summary = Yast::HTML.Para(summary)
+        summary = general_summary
 
-        # per zone overview
+        # per zone details
+        zones = firewalld.api.zones
         zones.each do |zone|
           summary << zone_summary(zone)
         end
@@ -244,6 +239,19 @@ module Y2Firewall
 
         summary = Yast::HTML.Heading(zone)
         summary << Yast::HTML.List(desc)
+      end
+
+      # Creates a general summary for firewalld
+      #
+      # @return [String] HTML formated firewall description
+      def general_summary
+        zones = firewalld.api.zones
+
+        summary = Yast::HTML.Bold("Default ZONE:") + " #{firewalld.api.default_zone}"
+        summary << Yast::HTML.Bold("Defined zones:")
+        summary << Yast::HTML.List(zones)
+
+        Yast::HTML.Para(summary)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


src/lib/y2firewall/clients/auto.rb
```diff
@@ -198,6 +206,33 @@ def imported
       def imported?
         !!self.class.imported
       end
+
+
+      FIREWALLD_ATTRS = [
```
teclator: It seems too complex for me:

Once the configuration has been read you don't need to use the `api` and will be more slow, and also the definition of the method to just one method in common is too much from my POV.

```ruby
firewalld.zones.each do |zone|
  summary << "<h2>zone.name</h2>"
  zone.relations.each do |relation|
    summary << "<p>#{relation.to_s.capitalize}: zone.public_send(relation).join(" ")</p>"
  end
```

teclator: BTW my minimal initial example was something like:

```ruby
summary << "<ul>"
summary << "<li><b>Default ZONE:</b> #{firewalld.default_zone}</li>"
summary << "<li><b>Log denied packets:</b> #{firewalld.log_denied_packets}</li>"
summary << "</ul>"
firewalld.zones.each do |z|
  summary << "<h2>#{z.name}</h2>"
  summary << "<p><b>Interfaces:</b> #{z.interfaces.join(',')}</p>"
  summary << "<p><b>Services:</b> #{z.services.join(',')}</p>"
  summary << "<p><b>Ports:</b> #{z.ports.join(',')}</p>"
  summary << "<p><b>Protocols:</b> #{z.protocols.join(',')}</p>"
  summary << "<p><b>Masquerade:</b> #{z.masquerade?}</p>"
end
```

mchf: I'll switch to upper layers soon. I've been focusing on layout so far.

src/lib/y2firewall/clients/auto.rb
```diff
@@ -198,6 +211,51 @@ def imported
       def imported?
         !!self.class.imported
       end
+
+      # Set of methods for formating all types of relations defined via
+      # has_many (@see Y2Firewall::Firewalld::Relations#has_many) in
+      # Y2Firewall::Firewalld::Zone
+      Firewalld::Zone.relations.each do |relation|
```
teclator: Still consider that using meta-programming for this is too much.

teclator:
@mchf if the logic grow too much I will prefer or would like to move it to a presenter.

src/lib/y2firewall/clients/auto.rb
```diff
@@ -198,6 +211,51 @@ def imported
...
+
+      # Creates a summary for the given zone
+      #
+      # @param [Firewalld::Zone] zone object defining a zone
+      # @return [String] HTML formated zone description
+      def zone_summary(zone)
+        raise ArgumentError, "zone parameter has to be defined" if zone.nil?
+
+        relations = Firewalld::Zone.relations.map(&:to_s)
+        desc = relations.map { |attr| send("#{attr}_summary", zone) }.delete_if(&:empty?)
```
teclator: You have a clear example here of what is discussed in the other PBI

src/lib/y2firewall/clients/auto.rb
mchf: can be, you know I was studying a bit and this looked as a simple enough example to give it a try in practice

src/lib/y2firewall/clients/auto.rb
mchf: hm, true. you got me ;-)

mchf:
well I'm more less done for now.

mchf:
so do you prefer adding a presenter now or later is enough?

teclator:
@mchf for the time being I think that does not make sense to move it to a presenter now but that is why I also preferred a simpler implementation

mchf:
yes tests are TBD as written in description. Methods are already private.

teclator:
@mchf: so all is fine and the life is wonderful :)

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -33,11 +33,70 @@
     allow(subject).to receive(:importer).and_return(importer)
   end
 
+  describe "#general_summary" do
```
teclator: I even does not expected this method to be tested as it is private but is ok

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -33,11 +33,70 @@
...
+    it "creates a text which includes firewalld overview" do
+      expect(firewalld).to receive(:running?).and_return(true)
+      expect(firewalld).to receive(:enabled?).and_return(true)
```
teclator: These calls shouldn't be expectations at all.

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -33,11 +33,70 @@
...
+      return nil if name.nil? || name.empty?
+      return nil if !Y2Firewall::Firewalld::Zone.instance_methods(false).include?(relation)
+
+      zone = Y2Firewall::Firewalld::Zone.new(name: name)
+      expect(zone).to receive(relation).and_return(values)
+
+      zone
+    end
+
+    it "empty zone returns empty description" do
```
teclator: what about using a context
```ruby
  context "when the zone does not have any relation"
    it "returns an empty description" do
    end
  end
```

or something similiar

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -33,11 +33,70 @@
...
+
+      expect(summary).to match(/Running/)
+      expect(summary).to match(/Enabled/)
+      expect(summary).to match(/Default zone/)
+      expect(summary).to match(/Defined zones/)
+    end
+  end
+
+  describe "#zone_summary" do
+    RELATIONS = {
```
teclator: I would use `let` to define the relations and the zones and a before loop if needed stubs with `allow`, I even would do at just testing the `summary` method allowing to call the originals of `zone_summary` and `general_summary` so you test them through the public method instead of from private calls

test/lib/y2firewall/clients/auto_test.rb
mchf: in fact I was thinking of it. But as I had just one test to put in I didn't do it. However, can do it so no problem.

test/lib/y2firewall/clients/auto_test.rb
mchf: why? the summary builds on to of results of these calls.

test/lib/y2firewall/clients/auto_test.rb
mchf: ok I'll mix all *summary tests together.

test/lib/y2firewall/clients/auto_test.rb
teclator: of course but it is not what u are testing, or is it?

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -33,52 +33,63 @@
...
+          protocols:  ["udp", "tcp"],
+          ports:      ["80"]
+        }
+      end
 
-      zone = Y2Firewall::Firewalld::Zone.new(name: name)
-      expect(zone).to receive(relation).and_return(values)
+      before(:each) do
+        zones = []
+        relations_stub.each_pair do |relation, values|
```
teclator: NP: You could even declare the zone with a let and then iterate over the stubs, and the relations can be assigned directly without mocking. Also as we are just checking that the summary contains the values, multiple zones are not necessary at all.

But I'm fine as it is.
But I'm fine as it is.

teclator:
Please update spec version and I will approve it.

PR #52, opened 2018-02-23
textdomain added for translation

schubi2:
Please review the following changes:
  * f16ff17 added textdomain


teclator:
@schubi2 thnx for taking care

PR #54, opened 2018-03-06
Use internal zone when the protection from INT is enabled

teclator:
When the profile contains the  `FW_PROTEC_FROM_INT` variable as "yes" then the `INT` zone will be mapped to the `internal` one.

Removed also the mapping of apache2 services into firewalld ones as the apache2 services definition will be finally shipped with the apache package.

Already documented in the AutoYaST main profiles changes from SLE12 to SLE15 PBI (https://github.com/yast/yast-autoinstallation/pull/408/files#diff-12924b8e025b11c36c24149f59c4be34R29)

### DIFF
--- a/test/lib/y2firewall/importer_strategies/suse_firewall.rb
+++ b/test/lib/y2firewall/importer_strategies/suse_firewall.rb
@@ -81,6 +81,16 @@ describe Y2Firewall::ImporterStrategies::SuseFirewall do
         expect(firewalld.default_zone).to eql("dmz")
       end
 
+      context "and protection from INT zone is not defined" do
+        let(:profile) { { "FW_DEV_INT" => "eth1" } }
+
+        it "configures the INT zone as the trusted" do
+          trusted = firewalld.find_zone("trusted")
+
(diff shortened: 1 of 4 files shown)


test/lib/y2firewall/importer_strategies/suse_firewall.rb
```diff
@@ -85,6 +81,24 @@
...
 
+      context "and protection from INT zone is disabled" do
+        it "configures the INT zone as the trusted" do
+          trusted = firewalld.find_zone("trusted")
+
+          expect(trusted.interfaces).to eq(["eth1"])
+        end
+      end
+
+      context "and protection from INT zone is enabled" do
```
jreidinger: maybe add also case when it is not defined at all?

teclator: ok, in that case it is like a disabled or not defined, but can add a separate case

jreidinger: I understand it from code, but it should be documented and tested in tests. so separate context make sense for me.

PR #57, opened 2018-03-23
Do not support FW_LOG_ACCEPT_CRIT tag

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1086655

It does not make sense to set the log of denied packets as "unicast" when the log of accepted critical packets is used. That is wrong, so fixed it.

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Mar 23 11:38:40 UTC 2018 - knut.anderssen@suse.com
+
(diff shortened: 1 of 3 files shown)


package/yast2-firewall.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Mar 23 11:38:40 UTC 2018 - knut.anderssen@suse.com
+
+- Do not support FW_LOG_ACCEP_CRIT tag as firewalld only log
```
mchf: typo + I don't understand the message anyway -> going to read it one more time ;-)

PR #58, opened 2018-03-27
Translate installation finish client title (bsc#1084136)

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1084136

### DIFF
--- a/src/lib/y2firewall/clients/installation_finish.rb
+++ b/src/lib/y2firewall/clients/installation_finish.rb
@@ -30,6 +30,8 @@ module Y2Firewall
     # the firewall proposal configuration for installation and autoinstallation
     # modes.
     class InstallationFinish < ::Installation::FinishClient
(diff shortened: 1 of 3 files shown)


src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -43,7 +45,7 @@ def initialize
       end
 
       def title
-        "Writing Firewall Configuration..."
+        _("Writing Firewall Configuration...")
```
teclator: The translation was available in "network" textdomain, but in this client we are using the firewall textdomain so not sure if it is ok, to just move the translation

lslezak: I think it's OK, just ping Stanislav Brabec, AFAIK he has some scripts which can move translations between textdomains.

PR #60, opened 2018-04-22
Installation: Use tigervnc firewalld service definition

teclator:
- [Trello Card](https://trello.com/c/WbNLhdQb/1414-5-15-ga-or-mu-bug-1088647-build-5502-openqa-test-fails-in-yast2vnc-firewalld-does-not-create-rule-for-vnc-and-bug-1088646-build)
- [bsc#1088646](https://bugzilla.suse.com/show_bug.cgi?id=1088646)

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Sat Apr 21 07:35:42 UTC 2018 - knut.anderssen@suse.com
+
+- AutoYaST: Use tigervnc firewalld service definition instead of
+  the firewalld one (bsc#1088646)
+- 4.0.24
+
+-------------------------------------------------------------------
 Tue Apr 17 11:55:47 CEST 2018 - snwint@suse.de
 
 - consistent wording (open/block) also for VNC ports (bsc#1089789)
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,8 +1,9 @@
 -------------------------------------------------------------------
 Sat Apr 21 07:35:42 UTC 2018 - knut.anderssen@suse.com
 
-- AutoYaST: Use tigervnc firewalld service definition instead of
-  the firewalld one (bsc#1088646)
+- During installation, open services defined by (Tigervnc) instead
+  of the 'vnc-server' service that is shipped with (firewalld)
+  (bsc#1081952).
 - 4.0.24
 
 -------------------------------------------------------------------
--- a/src/lib/y2firewall/clients/installation_finish.rb
+++ b/src/lib/y2firewall/clients/installation_finish.rb
@@ -54,9 +54,15 @@ module Y2Firewall
 
       def write
         Service.Enable("sshd") if @settings.enable_sshd
+        configure_firewall if @firewalld.installed?
+        true
+      end
 
-        return true if !@firewalld.installed?
+    private
 
+      # Modifies the configuration of the firewall according to the current
+      # settings
+      def configure_firewall
         @settings.enable_firewall ? @firewalld.enable! : @firewalld.disable!
 
         if @settings.open_ssh
@@ -65,9 +71,14 @@ module Y2Firewall
           @firewalld.api.remove_service(@settings.default_zone, "ssh")
         end
 
-        @firewalld.api.add_service(@settings.default_zone, "vnc-server") if @settings.open_vnc
-
-        true
+        if @settings.open_vnc
+          if @firewalld.api.service_supported?("tigervnc")
+            @firewalld.api.add_service(@settings.default_zone, "tigervnc")
+            @firewalld.api.add_service(@settings.default_zone, "tigervnc-https")
+          else
+            log.error "tigervnc service definition is not available"
+          end
+        end
       end
     end
   end
--- a/test/lib/y2firewall/clients/installation_finish_test.rb
+++ b/test/lib/y2firewall/clients/installation_finish_test.rb
@@ -6,6 +6,9 @@ require "y2firewall/clients/installation_finish"
 Yast.import "Service"
 
 describe Y2Firewall::Clients::InstallationFinish do
+  let(:proposal_settings) { Y2Firewall::ProposalSettings.instance }
+  let(:firewalld) { Y2Firewall::Firewalld.instance }
+
   describe "#title" do
     it "returns translated string" do
       expect(subject.title).to be_a(::String)
@@ -19,21 +22,13 @@ describe Y2Firewall::Clients::InstallationFinish do
   end
 
   describe "#write" do
-    let(:proposal_settings) { Y2Firewall::ProposalSettings.instance }
-    let(:api) do
-      instance_double(Y2Firewall::Firewalld::Api, remove_service: true, add_service: true)
-    end
-    let(:firewalld) { Y2Firewall::Firewalld.instance }
     let(:enable_sshd) { false }
-    let(:enable_firewall) { false }
     let(:installed) { true }
 
     before do
-      allow(proposal_settings).to receive("enable_sshd").and_return enable_sshd
-      allow(proposal_settings).to receive("enable_firewall").and_return enable_firewall
-      allow(firewalld).to receive("api").and_return api
-      allow(firewalld).to receive("installed?").and_return installed
-      allow(proposal_settings).to receive("open_ssh").and_return false
+      allow(proposal_settings).to receive("enable_sshd").and_return(enable_sshd)
+      allow(firewalld).to receive("installed?").and_return(installed)
+      allow(proposal_settings).to receive("open_ssh").and_return(false)
     end
 
     it "enables the sshd service if enabled in the proposal" do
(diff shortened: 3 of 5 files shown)


test/lib/y2firewall/clients/installation_finish_test.rb
```diff
@@ -47,48 +42,90 @@
...
+        expect(subject).to receive(:configure_firewall)
 
         subject.write
       end
 
-      it "disables the firewalld service if disabled in the proposal" do
-        expect(firewalld).to receive("disable!")
-
-        subject.write
+      it "returns true" do
```
mchf: I'd expect more verbose test description

test/lib/y2firewall/clients/installation_finish_test.rb
```diff
@@ -47,48 +42,90 @@
...
+          subject.send(:configure_firewall)
+        end
       end
 
-      it "returns true" do
-        expect(subject.write).to eq true
+      context "and the tigervnc service definition is not available" do
+        let(:service_available) { false }
+        it "logs the error" do
+          expect(subject.log).to receive(:error).with(/service definition is not available/)
```
mchf: logging is an additional tool for a programmer and as a such it should not be tested in my POV. Instead I'd expect test that the method returns ```false``` in this case

src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -54,9 +54,15 @@ def modes
 
       def write
         Service.Enable("sshd") if @settings.enable_sshd
+        configure_firewall if @firewalld.installed?
+        true
```
mchf: i think that methods which always returns ```true``` can easily be transformed into return ```nil```. No one will check the return value anyway.

On the other hand should it really be always true?

test/lib/y2firewall/clients/installation_finish_test.rb
imobachgs: IMHO it looks good. @teclator decided to write one "it" section for each thing he wants to test. If you check the previous one, you can read  *configures the firewall according to the proposal settings* (which is more verbose :smiley:). And this one is just about making sure that the method returns the right value.

mchf: if you imagine test overview, you'll get something like
```
When firewalld is installed
...
  returns true
```
and it doesn't look that good in my POV

src/lib/y2firewall/clients/installation_finish.rb
teclator: It will enable the ssh service and will configure the firewall according to the proposal. Once done it will return true, if something fails with the firewalld configuration, it will be reported by the firewalld command line, except for the support of the service that will be logged.

I decided to return true once the client did its job, in case in the future we would like to return false in some condition we can add it safety.

And it is a common return implementation over finish clients:

https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/finish_client.rb#L73

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: It is an additional tool for a programmer but it is also part of the program and the workflow and it describes what the client does when the tigervnc service definition is not present.

It could raise an error, report an error or whatever. In this case it just logs de error.

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: Exactly ;)

teclator: @imobachgs exactly

src/lib/y2firewall/clients/installation_finish.rb
mchf: yes, I can read the code.

The point should be that it is already masking return value of ```Service.Enable```. And returning true when firewall could not be configured is arguable but acceptable in the finish client.

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: @mchf and it describes explicitly what the methods returns without confusion. I decided to put inside the context because from my POV it describes better that without firewalld installed it does nothing special and shows that the scenarios is covered.

So, which description would you prefer?:

```
#write
  enables the sshd service if enabled in the proposal
  when firewalld is not installed
    returns true
  when firewalld is installed
    configures the firewall according to the proposal settings
    returns true
```

```
#write
  enables the sshd service if enabled in the proposal
  returns true
  when firewalld is installed
    configures the firewall according to the proposal settings
```

test/lib/y2firewall/clients/installation_finish_test.rb
mchf: well I'm still not convinced. Logging is for programmer to get some meta information about the algorithm. However workflow should be driven by return values, exceptions, ... and these things should be covered.

It e.g. happens quite often to me that e.g. when debugging something I'm shifting, disabling and modifying log messages quite often. Each of these actions can influence the test. In the best case it will only break building the test package.

test/lib/y2firewall/clients/installation_finish_test.rb
mchf: the first one ;-) But anyway this is NP for me.

src/lib/y2firewall/clients/installation_finish.rb
teclator: If I'm not wrong in other clients it was not a condition of return. I mean, we tried to enable the service failing silently and it behaves exactly as it was.

https://github.com/yast/yast-network/blob/SLE-12-SP1/src/clients/firewall_stage1_finish.rb#L68

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: So it seems there are two different things... whether it should be tested or not and whether we should fail just logging the error or reporting it.

Again will use an example from the same client that also logs the error instead of raise exceptions:

https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/finish_client.rb#L91

Of course in our scenario the package should be present, and well I can omit the check of supported service just trying to enable the services and relying in the firewalld API for the error reporting or just add a report instead of the log if we think it deserve it.

About the test, as commented we are testing all the possible scenarios and we are ensuring we are logging it and it even takes more importance if as you say you modify the log messages or disable whatever. It is not the first time we do a typo logging and error or a warning and the customer founds the error first than us just because we didn't checked all the possible scenarios.

imobachgs: I do not change logs texts that often, but when I do fixing the test in those cases is a no-brainer. Anyway, there is another important benefit IMHO: prevent introducing typos. Let's say that, in the future, I change this `log.error` for `log.warning`; or I introduce some interpolation in the string (and I do it wrong)... In those cases, I am covered and YaST won't crash.

src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -65,9 +71,14 @@ def write
           @firewalld.api.remove_service(@settings.default_zone, "ssh")
         end
 
-        @firewalld.api.add_service(@settings.default_zone, "vnc-server") if @settings.open_vnc
-
-        true
+        if @settings.open_vnc
+          if @firewalld.api.service_supported?("tigervnc")
```
teclator: From my POV, the most important discussion that should be solve is what to do if the service is not defined during the installation. The [package](https://github.com/yast/yast-packager/pull/286/files#diff-0b92eb80e55af7a47d095f459c8ac24cR21) that ships the service definition should be installed (verified manually in SLE-15 and also verified the behavior before firewalld in SLE-12-SP3)

If for some reason it is not supported during the installation we just log an error. We could report an error instead or just do not check it and rely on firewalld api for the error reporting.

package/yast2-firewall.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Sat Apr 21 07:35:42 UTC 2018 - knut.anderssen@suse.com
+
+- AutoYaST: Use tigervnc firewalld service definition instead of
```
mvidner: AFAICT this bug is not specific to AutoYaST.

Also, the changelog wording suggests that there are two service definitions one of which is wrong so we are using the other one. But in [Bug 1088647](https://bugzilla.suse.com/show_bug.cgi?id=1088647) you say:
> The service is not called vnc-server anymore but tigervnc and tigervnc-https instead.

So the changelog should say something like this, right?
> Use the correct service name (tigervnc) for opening the VNC port in the firewall

src/lib/y2firewall/clients/installation_finish.rb
mvidner: I think you are both missing an important point: you are not designing an API. You are implementing an instance of the [FinishClient](http://www.rubydoc.info/github/yast/yast-yast2/Installation/FinishClient) API. So the **caller** decides how success and failure is communicated.

(Unfortunately, [the API docs is silent on this point](https://github.com/yast/yast-yast2/blob/f829c95f3f284fff9ee93a77d3278fb2ecffe942/library/general/src/lib/installation/finish_client.rb#L66))

teclator: Well, it maybe came from ycp times but that extra checks are usually not done:

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/clients/finish.rb

teclator: Maybe something to document? Best practices for installation finish clients?

teclator:
@mvidner adapted the changelog, and also used the `bug id` where was decided to use services shipped by **Tigervnc** instead of the 'vnc-server' definition that is shipped with **firewalld**

PR #61, opened 2018-04-26
Checking semantic while importing AY settings

schubi2:


src/lib/y2firewall/clients/auto.rb
```diff
@@ -88,6 +89,18 @@ def import(profile)
         enable if profile.fetch("enable_firewall", settings.enable_firewall)
         start if profile.fetch("start_firewall", false)
         importer.import(profile)
+
+        # Checking if an interface has been defined for different zones
```
kobliha: Nitpicking: we expect this to be extended in the future, a separate function, e.g., `check_profile_for_errors` might be better

PR #63, opened 2018-05-04
added build requirments

schubi2:


package/yast2-firewall.spec
```diff
@@ -27,6 +27,8 @@ Group:	        System/YaST
 License:        GPL-2.0
 BuildRequires:  perl-XML-Writer update-desktop-files yast2-testsuite
 BuildRequires:  yast2-devtools >= 3.1.10
+# running testcases while build
+BuildRequires:  autoyast2-installation
```
lslezak: Hm, I'm not sure if this is a good idea. This will make the build chain longer...

lslezak: See https://github.com/yast/yast-installation/blob/master/test/test_helper.rb#L19-L36

PR #70, opened 2018-08-20
Changed licence to GPL-2.0-only.

schubi2:


jreidinger:
travis failing

package/yast2-firewall.spec
```diff
@@ -64,6 +64,11 @@ rake test:unit
 %install
 rake install DESTDIR="%{buildroot}"
 
+# Remove the license from the /usr/share/doc/packages directory,
+# it is also included in the /usr/share/licenses directory by using
+# the %license tag.
+rm $RPM_BUILD_ROOT/%{yast_docdir}/COPYING
```
jreidinger: something that should be fixed at https://github.com/yast/yast-rake/blob/master/lib/tasks/install.rake#L54 by removing COPYING from it.

PR #75, opened 2018-09-10
Read-write ports, via PortsForProtocols

mvidner:


src/lib/y2firewall/widgets/pages/zone.rb
```diff
@@ -72,60 +74,92 @@ def label
...
+          # @param hash [Hash{Symbol => Array<String>}] ports specification broken by protocol
+          # @example
+          #   h = { tcp: ["55555-55666", "44444"], udp: ["33333"] }
+          #   a = ["55555-55666/tcp", "44444/tcp", "33333/udp"]
+          #   ports_to_array(h) # => a
+          def ports_to_array(hash)
+            hash.map { |sym, ports| ports.map { |p| "#{p}/#{sym}" } }.flatten
+          end
+
+          # @param s [String] ports specification as one string
```
imobachgs: param is called `a` instead of `s`. Moreover, I would prefer a more descriptive name :)

PR #79, opened 2018-09-14
Features/startup

teclator:
It depends on https://github.com/yast/yast-yast2/pull/827


src/lib/y2firewall/widgets/pages/startup.rb
```diff
@@ -29,30 +30,51 @@ module Pages
       # A page for firewall service startup
       class Startup < CWM::Page
         # Constructor
-        #
-        # @param pager [CWM::TreePager]
-        def initialize(_pager)
+        def initialize
```
mvidner: The caller in overview.rb has not been fixed to omit the argument

src/lib/y2firewall/dialogs/main.rb
```diff
@@ -39,7 +39,8 @@ def initialize
       end
 
       def title
-        _("Firewall")
+        # FIXME: the api mode is temporal, should be removed
```
imobachgs: Do you need to show the `mode` here? Is it translated?

src/lib/y2firewall/dialogs/main.rb
```diff
@@ -58,9 +59,12 @@ def run
 
         loop do
           result = super
-          break unless result == :redraw
+          swap_api if result == :swap_mode
+          break unless result == :redraw || result == :swap_mode
         end
 
+        fw.write_only
```
imobachgs: Are you writing the changes *always*? What if the user uses the `Abort` button?

src/lib/y2firewall/dialogs/main.rb
```diff
@@ -100,6 +100,10 @@ def back_handler
       def fw
         Y2Firewall::Firewalld.instance
       end
+
+      def swap_api
```
imobachgs: I would like to see some description here. Why are you assigning a new API object? Why? (I know why, but it would be nice to document it).

src/lib/y2firewall/clients/firewall_new.rb
```diff
@@ -33,12 +33,18 @@ class FirewallNew
       include Yast::I18n
       include Yast::Logger
 
+      # Constructor
       def initialize
         textdomain "firewall"
+
+        Yast.import "UI"
```
imobachgs: Unless there is a strong reason against that, you should move `import` calls next to regular Ruby `require` calls. Actually, they are mostly *requires*.

src/lib/y2firewall/widgets/overview.rb
```diff
@@ -75,7 +75,7 @@ def items
 
       # @return [CWM::PagerTreeItem]
       def startup_item
-        page = Pages::Startup.new(self)
+        page = Pages::Startup.new
```
imobachgs: I am fine with removing the `_pager` argument. But we should be consistent and do the same for the rest of pages in this module.

test/lib/y2firewall/dialogs/main_test.rb
```diff
@@ -54,12 +54,26 @@
...
     end
 
     context "when the user accepts the changes" do
+      it "writes the firewall configuration" do
+        expect(firewall).to receive(:write_only)
+
+        subject.run
+      end
+
+      it "updates the firewalld systemd service status" do
```
imobachgs: I would write another test to be sure that the `save` method *is not called* when the user aborts.

src/lib/y2firewall/dialogs/main.rb
teclator: Probably yes, when we moved it from the next_handler it was not checked whether the result was `:next`or not

src/lib/y2firewall/dialogs/main.rb
teclator: No, we do not need it

PR #80, opened 2018-09-17
Validate the Ports tab input

mvidner:


src/lib/y2firewall/widgets/pages/zone.rb
```diff
@@ -119,16 +129,37 @@ def init
...
-          # FIXME: validation, cleanup, error reporting
           def store
-            by_proto = PROTOCOLS.map do |sym, _label|
+            by_proto = values
+            @zone.ports = ports_to_array(by_proto.to_h)
+          end
+
+          def validate
+            by_proto = values
+            by_proto.each do |sym, ranges|
```
imobachgs: np: Perhaps I would change `sym` by `proto` to make it clear.

PR #81, opened 2018-09-17
Add a list of interfaces

imobachgs:
This PR implements support to assign interfaces to firewall zones.

change-zone-interface


src/lib/y2firewall/dialogs/change_zone.rb
```diff
@@ -0,0 +1,122 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm/dialog"
+require "y2firewall/widgets/zone_options"
+
+module Y2Firewall
+  module Dialogs
+    # This dialog allows the user to select which zone should be an interface assigned to.
```
teclator: So, do you prefer a normal dialog instead of a Popup one? IMHO it looks like too much space for just a combo box but feel free to ignore this comment.

src/lib/y2firewall/ui_state.rb
```diff
@@ -0,0 +1,155 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Firewall
+  # Singleton class to keep the position of the user in the UI and other similar
+  # information that needs to be rememberd across UI redraws to give the user a
+  # sense of continuity.
+  class UIState
```
mvidner: Something like this is also used in the Partitioner. Is it easy or hard to extract the generic part?

src/lib/y2firewall/ui_state.rb
```diff
@@ -0,0 +1,155 @@
...
+    #
+    # @param page [CWM::Page]
+    # @param candidate [Integer, String]
+    # @return boolean
+    def matches?(page, candidate)
+      page.label == candidate
+    end
+
+    class << self
+      # Singleton instance
```
mvidner: Why not use the stdlib https://devdocs.io/ruby~2.5/singleton ?

src/lib/y2firewall/ui_state.rb
```diff
@@ -0,0 +1,155 @@
...
+    # @param [CWM::Page] page associated to the tab
+    def switch_to_tab(page)
+      self.tab = page.label
+    end
+
+    # Method to be called when the user operates in a row of a table of zones
+    # or creates a new zone.
+    #
+    # @param zone [String] zone name
+    def select_row(zone_name)
```
teclator: As we were discussing we should try to generalize the current. But for now we should at least fix the comments and param name as it will used not only for zones but also for interfaces.

src/lib/y2firewall/ui_state.rb
```diff
@@ -0,0 +1,155 @@
...
+    attr_reader :tab
+    # @see #tab
+    def tab=(tab)
+      @tab = tab
+      # If the user switched to a new tab, invalidate details about the inner
+      # table
+      self.row_id = nil
+    end
+
+    # List of candidate nodes to go back after opening a zone view in the tree
```
teclator: Same than above

src/lib/y2firewall/widgets/interfaces_table.rb
```diff
@@ -0,0 +1,98 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm"
+require "cwm/table"
+require "y2firewall/ui_state"
+
+module Y2Firewall
+  module Widgets
+    # A table with all {Y2Firewall::Firewalld::Zone}s.
```
mvidner: not zones, interfaces

teclator:
Just some minor comments.

src/lib/y2firewall/widgets/interfaces_table.rb
```diff
@@ -0,0 +1,98 @@
...
+    class InterfacesTable < ::CWM::Table
+      DEFAULT_ZONE_NAME = "default".freeze
+
+      # @!attribute [r] interfaces
+      #   @return [Array<Hash>] Interfaces
```
mvidner: When a Hash is really an ad-hoc class, like it is here for the interface, I like to make a subclass of Hash just for documentation purposes, like here:
declaration: https://github.com/yast/yast-yast2/blob/e4ab10035ba30f739175009344872c4062c26db6/library/systemd/src/lib/yast2/systemd/unit_prop_map.rb#L8
reference: https://github.com/yast/yast-yast2/blob/e4ab10035ba30f739175009344872c4062c26db6/library/systemd/src/lib/yast2/systemd/socket.rb#L70

So I think we should declare Y2Firewall::Firewalld::Interface

src/lib/y2firewall/ui_state.rb
imobachgs: Just a leftover. Anyway, I will try to extract it.

src/lib/y2firewall/ui_state.rb
imobachgs: Yes, that's true. I took the adapted code from @teclator (thanks!) and I will try to extract the common pieces and move it to yast2.

src/lib/y2firewall/ui_state.rb
imobachgs: I do not know, this code comes from yast2-storage-ng. But I do not see a good reason.

src/lib/y2firewall/widgets/interfaces_table.rb
imobachgs: I agree. @teclator please, could you adapt `known_interfaces` to work with proper objects while I am extracting the popup and the `UIState` classes?

teclator: @mvidner @imobachgs please see https://github.com/yast/yast-yast2/pull/832

src/lib/y2firewall/widgets/interfaces_table.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      # @macro seeAbstractWidget
+      def opt
+        [:notify, :immediate]
+      end
+
+      # @macro seeAbstractWidget
+      def init
+        interface = Y2Firewall::UIState.instance.row_id
+        if interface && interfaces.map(&:name).include?(interface.to_s)
```
teclator: np: why the `id` is not used?

```ruby
interfaces.map(&:id).include?(interface)
```

src/lib/y2firewall/widgets/interfaces_table.rb
```diff
@@ -0,0 +1,100 @@
...
+          _("Zone"),
+          _("Name")
+        ]
+      end
+
+      # @see CWM::Table#items
+      def items
+        interfaces.map do |iface|
+          [
+            iface.name.to_sym,
```
teclator: `iface.id`?

src/lib/y2firewall/widgets/interfaces_table.rb
```diff
@@ -0,0 +1,100 @@
...
+      # @macro seeAbstractWidget
+      def handle(event)
+        return nil unless event["EventReason"] == "SelectionChanged"
+        UIState.instance.select_row(value)
+        change_zone_button.interface = selected_interface
+        nil
+      end
+
+      def selected_interface
+        interfaces.find { |i| i.name == value.to_s }
```
teclator: Same than above

src/lib/y2firewall/widgets/zone_options.rb
```diff
@@ -0,0 +1,89 @@
...
+      #
+      # @param interface [Y2Firewall::Firewalld::Interface] Interface to act on
+      def initialize(interface)
+        textdomain "firewall"
+        @interface = interface
+      end
+
+      # @macro seeAbstractWidget
+      def init
+        return unless @interface.zone
```
teclator: np: you can use the getter method

src/lib/y2firewall/widgets/zone_options.rb
```diff
@@ -0,0 +1,89 @@
...
+        [DEFAULT_ZONE_OPTION] + zones.map { |z| [z.name, z.name] }
+      end
+
+      # @macro seeCommonWidget
+      def store
+        new_zone = selected_zone
+        return if new_zone && new_zone.interfaces.include?(interface.name)
+
+        old_zone = interface.zone
+        old_zone.remove_interface(interface.name) if old_zone
```
teclator: Maybe a change zone method could be added to the interface object (for SP1 :roll_eyes: )

teclator:
LGTM

src/lib/y2firewall/widgets/zone_options.rb
imobachgs: Yes, it would be nice :)

PR #85, opened 2018-09-20
Allow to change the default zone

imobachgs:
This PR allows to change the Firewalld's default zone.

change-default-zone

### DIFF
--- a/src/lib/y2firewall/widgets/zones_table.rb
+++ b/src/lib/y2firewall/widgets/zones_table.rb
@@ -32,22 +32,74 @@ module Y2Firewall
       # Constructor
       #
       # @param zones [Array<Y2Firewall::Firewalld::Zone>] Zones
-      def initialize(zones)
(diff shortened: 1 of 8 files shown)


src/lib/y2firewall/widgets/zones_table.rb
```diff
@@ -32,22 +32,74 @@ class ZonesTable < ::CWM::Table
...
+        default_zone_button.zone = selected_zone
       end
 
       # @see CWM::Table#header
       def header
         [
           _("Name"),
-          _("Interfaces")
+          _("Interfaces"),
+          _("Default")
```
teclator: I'm not sure about having an extra column for it, but probably that is something for Ken

imobachgs: We have something similar in the repository list. But we might append a "(default)" string if preferred.

teclator:
LGTM once travis is fine with the rdoc check :wink:

imobachgs:
> LGTM once travis is fine with the rdoc check wink

Thanks :)

PR #87, opened 2018-09-20
Display interfaces which are assigned to the 'default' zone

imobachgs:
set-default-zone

### DIFF
--- a/src/lib/y2firewall/widgets/zones_table.rb
+++ b/src/lib/y2firewall/widgets/zones_table.rb
@@ -108,7 +108,7 @@ module Y2Firewall
       def assigned_interfaces(zone)
         assigned = zone.interfaces.clone
         assigned += interfaces.reject(&:zone).map(&:name) if default_zone?(zone)
-        assigned
+        assigned.sort
       end
 
       # Returns the default zone


teclator:
Well, not sure about this fix, the interfaces are not assigned explicitly.

imobachgs:
> Well, not sure about this fix as the interfaces are not assigned explicitly and we are working with the permanent configuration.

Sorry, I am not sure about what you are talking about. The idea is that, when you assign an interface to the "Default" zone, I would expect it to be changed when you change the "Default" one.

For instance, you assign `eth0` to the `Default` zone which is `public`. Then, you set `home` as the default zone. I would expect `eth0` to be assigned them to `home` as it is now the default zone.

But most probably I am missing some detail.

teclator:
That is completely right. What I'm saying is that them are not assigning explicitly to the "home" zone but to the default or not zone at all.

imobachgs:
Yes, and as a user, I expect those interfaces to listed there. But I see your point now. Perhaps we could add some indication so the user knows that those interfaces are there because they were assigned to the default zone.

teclator:
Something to be discussed with Ken, and probably for SP1 once we get some feedback. Maybe we could add in the help each zone description, but that is just an idea.

PR #90, opened 2018-09-21
Remove completely old susefirewall dialogs and modules

teclator:


teclator:
@imobachgs @mvidner It seems it is not used in any other place, so we could remove it already in `SLE-15-GA`, and remove the rest (yast2 and yast2-network modules) of the old `SuSEFirewall2` code only from the `master` branch

imobachgs:
If it is not used, I am fine with removing it. But I am not sure if it is a good thing for a MU.

teclator:
Yep, that was my main doubt and still it is :yum:

teclator:
@jreidinger I resolved the current conflicts, the PR was already approved although would be nice to have a final LGTM

teclator:
@schubi2 thnx!

PR #92, opened 2018-10-02
Add firewalld autoyast configuration support.

teclator:
- [Trello card](https://trello.com/c/afBwY7Rc/345-sle-15-ga-as-mu-firewall-ui-support-for-autoyast)
- It depends on yast/yast-yast2#840

### AutoYaST Profile (only firewall section)

```xml
  <firewall>
    <zones config:type="list">
      <zone>
        <name>public</name>
        <interfaces config:type="list">
          <interface>eth0</interface>
        </interfaces>
        <services config:type="list">
          <service>ssh</service>
          <service>http</service>
          <service>https</service>
          <service>vnc</service>
        </services>
        <ports config:type="list">
          <port>8080/tcp</port>
          <port>9090/udp</port>
        </ports>
      </zone>
      <zone>
        <name>home</name>
        <services config:type="list">
          <service>ssh</service>
        </services>
        <ports config:type="list">
          <port>22/tcp</port>
          <port>5353/udp</port>
        </ports>
      </zone>
    </zones>
    <default_zone>public</default_zone>
    <log_denied_packets>off</log_denied_packets>
    <enable_firewall config:type="boolean">true</enable_firewall>
    <start_firewall config:type="boolean">true</start_firewall>
  </firewall>
```

### Difference with the defaults

```diff
diff --git a/autoinst.xml b/autoinst.xml
index ca22ff6..e44775b 100644
--- a/autoinst.xml
+++ b/autoinst.xml
@@ -64,13 +64,13 @@
         <interfaces config:type="list"/>
         <masquerade config:type="boolean">false</masquerade>
         <name>home</name>
-        <ports config:type="list"/>
+        <ports config:type="list">
+          <port>22/tcp</port>
+          <port>5353/udp</port>
+        </ports>
         <protocols config:type="list"/>
         <services config:type="list">
           <service>ssh</service>
-          <service>mdns</service>
-          <service>samba-client</service>
-          <service>dhcpv6-client</service>
         </services>
         <short>Home</short>
         <sources config:type="list"/>
@@ -95,14 +95,20 @@
       </zone>
       <zone>
         <description>For use in public areas. You do not trust the other computers on networks to not harm your computer. Only selected incoming connections are accepted.</description>
-        <interfaces config:type="list"/>
+        <interfaces config:type="list">
+          <interface>eth0</interface>
+        </interfaces>
         <masquerade config:type="boolean">false</masquerade>
         <name>public</name>
-        <ports config:type="list"/>
+        <ports config:type="list">
+          <port>8080/tcp</port>
+          <port>9090/udp</port>
+        </ports>
         <protocols config:type="list"/>
         <services config:type="list">
           <service>ssh</service>
-          <service>dhcpv6-client</service>
+          <service>http</service>
+          <service>https</service>
         </services>
         <short>Public</short>
         <sources config:type="list"/>
```


src/lib/y2firewall/clients/auto.rb
```diff
@@ -62,28 +67,20 @@ def initialize
...
 
       # Import the firewall configuration
       #
       # @param profile [Hash] firewall profile section to be imported
       # @return [Boolean]
-      def import(profile)
+      def import(profile, merge = !Yast::Mode.config)
         self.class.profile = profile
-        return false unless read
+        return false if merge && !read(force: false)
```
schubi2: Hm, do you have tested this while a normal AY installation too ? With which configuration it will be merged ?
(It could depend on if we are in first or second stage)

teclator: In a normal AutoYaST installation, we are always trying to merge the configuration with the current one avoiding to read it twice. 

Firewalld ships with a default configuration, predefined zones and services, so we read that configuration and just apply changes over it.

One of the problem with AutoYaST configuration is that we do some imports and exports when editing, so we need to avoid a read of the configuration in that cases, we only need to read when `Applying to the system`, and in this case we should remember the state before writing not showing the current system configuration but only the changes.

teclator: @schubi2 
Tested it, also take into account that currently the firewall finish client does not support autoinstallation so, only called during a normal installation. That means that we are configuring the firewall only in the second stage.

PR #93, opened 2018-10-02
AutoYaST schema: New zone attributes (bsc#1108199)

teclator:
In this [commit](https://github.com/yast/yast-yast2/pull/818/files#diff-1b7047b35ed0976769eccd6200a38d82R56) we added support for more zone attributes  but we forgot to update the firewall schema.

This PR fix it.

- https://bugzilla.suse.com/show_bug.cgi?id=1108199

### DIFF
--- a/src/autoyast-rnc/firewall.rnc
+++ b/src/autoyast-rnc/firewall.rnc
@@ -126,6 +126,9 @@ zones =
     LIST,
     element (zone | listentry) {
       zone_name &
+      zone_short? &
+      zone_description? &
+      zone_target? &
       fwd_interfaces? &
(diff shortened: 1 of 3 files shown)


src/autoyast-rnc/firewall.rnc
```diff
@@ -126,6 +126,9 @@ zones =
     LIST,
     element (zone | listentry) {
       zone_name &
+      zone_short &
```
imobachgs: Oh, wait, are those attributes mandatory?

teclator:
@imobachgs good catch, thnx

PR #96, opened 2018-10-09
Warn the user about unsaved changes (services selection)

teclator:
- Trello Card: https://trello.com/c/afBwY7Rc/345-sle-15-ga-as-mu-firewall-ui-support-for-autoyast
## UX improvements

screenshot from 2018-10-09 12-32-48

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct  9 08:55:56 UTC 2018 - knut.anderssen@suse.com
+
+- Alert the user about unsaved changes when leaving the zone
+  services configuration without applying changes (fate#324662)
+- 4.0.30
+
+-------------------------------------------------------------------
 Fri Oct  5 06:49:03 UTC 2018 - knut.anderssen@suse.com
 
 - AutoYaST configuration: Added support to editing and reseting the
--- a/package/yast2-firewall.spec
+++ b/package/yast2-firewall.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        4.0.29
+Version:        4.0.30
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/src/lib/y2firewall/widgets/allowed_services.rb
+++ b/src/lib/y2firewall/widgets/allowed_services.rb
@@ -92,6 +92,15 @@ module Y2Firewall
         )
       end
 
+      def validate
+        return true if selected_services.empty?
+        # TRANSLATORS: popup question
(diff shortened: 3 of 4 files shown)


imobachgs:
np: I am not sure about the term "applying".

teclator:
> IMHO people will confuse the "applying" with "accepting" the whole module. I would call it "adding", like the button that does it.

That is I why I specified "the selection of services", and it is for both actions ("adding" or "removing" services")
 
> 
> Also, we are giving 2 choices:
> 1: Yes = Switch tabs, discarding the selected services
> 2: No = Don't switch tabs
> 
> But most likely, the user wants to do something different:
> 3: Switch tabs _and_ Add the services to the Allowed set beforehand.

Yep, I was thinking on this option also, but discarded for not complicate it more.

> 
> So I would like to have a choice between 3 and 1, omitting 2 entirely:
> 
> > The selected services have not yet been added to the allowed set
> > [Add] [Discard]

In this case, [Apply Changes][Discard] ?

mvidner:
Ah, you mean [Apply Changes] means to Add the ones in the left column **and** Remove those in the right column, right? I missed that because the screenshot does not show that complex case.

Well, then  [Apply Changes][Discard] is good enough. (We could make it clearer but the code would get too complicated)

teclator:
> Ah, you mean [Apply Changes] means to Add the ones in the left column **and** Remove those in the right column, right? I missed that because the screenshot does not show that complex case.

Yep, I realized that and that is why I have updated the screenshot, hope it is more clear now.
> 
> Well, then [Apply Changes][Discard] is good enough. (We could make it clearer but the code would get too complicated)

Exactly, and in this case I am not sure if we should summarize which services will be added and which ones will be removed.

teclator:
It seems that [Apply Changes] could be even a nive improvement, as currently you are only allowed to submit one of the list changes (`Add` or `Remove`)

imobachgs:
> It seems that [Apply Changes] could be even a nive improvement, as currently you are only allowed to submit one of the list changes (`Add` or `Remove`)

Take into account that we took inspiration from the partitioner, so we should keep the same approach (to keep consistency).

teclator:
> > It seems that [Apply Changes] could be even a nive improvement, as currently you are only allowed to submit one of the list changes (`Add` or `Remove`)
> 
> Take into account that we took inspiration from the partitioner, so we should keep the same approach (to keep consistency).

So, what do you suggest?

teclator:
@imobachgs @mvidner mmm, what about leaving it as it is until we get some feedback from @kwwii

imobachgs:
> So, what do you suggest?

To be honest, I do not have a better alternative. I only mentioned that the "apply" term could be confusing.

PR #103, opened 2018-11-02
Enable and open SSH port when authentication is done via public key

imobachgs:
When no password but an SSH public key has been specified in the installation for the root user,
YaST proposes to open the SSH service in order to allow logging into the system.

See https://github.com/yast/yast-users/pull/173

### DIFF
--- a/package/yast2-firewall.spec
+++ b/package/yast2-firewall.spec
@@ -32,13 +32,9 @@ BuildRequires:  yast2-devtools >= 3.1.10
 BuildRequires:  yast2 >= 4.1.21
 BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
 BuildRequires:  rubygem(%rb_default_ruby_abi:rspec)
-# Yast::UsersSimple.GetRootPassword
-BuildRequires:  yast2-users
 
 # Y2Firewall::Firewalld#reset
 Requires:       yast2 >= 4.1.21
-# Yast::UsersSimple.GetRootPassword
-Requires:       yast2-users
 
 # ButtonBox widget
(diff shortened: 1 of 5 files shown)


package/yast2-firewall.spec
```diff
@@ -32,9 +32,13 @@ BuildRequires:  yast2-devtools >= 3.1.10
 BuildRequires:  yast2 >= 4.1.21
 BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
 BuildRequires:  rubygem(%rb_default_ruby_abi:rspec)
+# Yast::UsersSimple.GetRootPassword
+BuildRequires:  yast2-users
 
 # Y2Firewall::Firewalld#reset
 Requires:       yast2 >= 4.1.21
+# Yast::UsersSimple.GetRootPassword
+Requires:       yast2-users
```
kobliha: I'm not sure if we actually want this. It may create some other dependency cycle. At the end, it's not a runtime dependency, just in Installer.

package/yast2-firewall.spec
```diff
@@ -32,9 +32,13 @@ BuildRequires:  yast2-devtools >= 3.1.10
 BuildRequires:  yast2 >= 4.1.21
 BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
 BuildRequires:  rubygem(%rb_default_ruby_abi:rspec)
+# Yast::UsersSimple.GetRootPassword
+BuildRequires:  yast2-users
```
kobliha: Maybe it's really needed. Ideally we don't create such new dependencies :-/

PR #109, opened 2018-12-20
Startup dialog: Use the new CWM::ServiceWidget (bsc#1114807)

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1114807

### DIFF
--- a/src/lib/y2firewall/widgets/pages/startup.rb
+++ b/src/lib/y2firewall/widgets/pages/startup.rb
@@ -21,8 +21,8 @@
 
 require "yast"
 require "cwm/page"
+require "cwm/service_widget"
 require "y2firewall/firewalld"
-require "ui/service_status"
 
 module Y2Firewall
   module Widgets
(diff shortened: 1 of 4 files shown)


src/lib/y2firewall/widgets/pages/startup.rb
```diff
@@ -44,35 +44,17 @@ def label
...
-        def help
-          status_widget.help
-        end
-
       private
 
         # This is a generic widget in SLE15; may not be appropriate.
         # For SLE15-SP1, use CWM::ServiceWidget
         def status_widget
-          @status_widget ||= ::UI::ServiceStatus.new(system_service.service)
```
jreidinger: method comment looks outdated

teclator: Yep, that is true. @dgdavid if you plan to improve it then take this into account ;)

dgdavid: Yes, I noticed it and I take it in mind. Anyway, thank you both, for the comment and the ping. That way, I shoul not forgot it ;)

PR #113, opened 2019-01-08
Master 1114673

schubi2:


src/lib/y2firewall/dialogs/main.rb
```diff
@@ -130,6 +130,13 @@ def swap_api
       # modifications
       def apply_changes
         return false if Yast::Mode.config
+        # Firewall settings will be written into the permanent configurations only.
+        # So the running firewalld service will not be changed. Even a reload does
+        # not help (see man pages). So the running firewalld service has to be
```
teclator: Well, the man is talking about the Runtime configuration, isn't it?, so writing to runtime and doing a reload will lost the changes but doing it in the permanent configuration and doing a reload should work fine.

teclator: Permanent configuration
       The permanent configuration is stored in config files and will be
       loaded and become new runtime configuration with every machine boot or
       service reload/restart.

schubi2: It depends. While writing the settings we are using --permanent. So the settings will be written to permanent  configuration only. After restarting the service it will affect the running configuration.

src/lib/y2firewall/dialogs/main.rb
```diff
@@ -130,6 +130,17 @@ def swap_api
...
+        # restarted.
+        # Set a flag only. Restarting will be done by system_service.save.
+        if fw.modified? && # Data has been changed by user
+            fw.system_service.running? &&      # The service is already running
+            fw.system_service.action != :stop  # and will not be stopped by the user
```
teclator: Still not sure at all, I mean, if the user select a reload after writing we will restart the service that isn't what was requested at all.

schubi2: I see your point. Well, from my point of view I would not offer "restart" or "reload" at all to the user, because even for us it is not really clear in all corner cases. So I would say...keep it simple for the user. What do you think about it ?

teclator: Yep, that are the consequences of moving to the service_widget, it is a general approach while in the past it was handled by the firewall itself.

BTW, the only problematic case is when the service is `running` and the configuration has been modified, in this case a firewalld --reload should be enough what is basically what `fw.write` does.

So maybe something like

```ruby
... 
  return fw.write if need_reload?

  fw.write_only
  fw.system_service.save
end

def need_reload?
  fw.modified? && fw.system_service.running? && !fw.system_service.action
end

```

schubi2:
So Knut and I have come to the conclusion that we take this PR until firewalld is more "clean" here.

src/lib/y2firewall/dialogs/main.rb
teclator: @schubi2 

**Note:** this is the bug about --reload that I was talking about https://bugzilla.opensuse.org/show_bug.cgi?id=1112008

PR #116, opened 2019-01-14
require yast2 >= 3.2.49 to ensure we have firewall_chooser (bsc#1121627)

wfeldt:
Note that this version number is *different* than for SLE12-SP3. What a
versioning mess. :-(


teclator:
@wfeldt thnx for taking care, and next time I will not forgot to increase the minor version when diverging.

PR #117, opened 2019-01-14
Autoyast adaptation

jreidinger:
trello: https://trello.com/c/5OWTqVV4/594-fate-324662-sle-15-sp1-yast-module-for-firewalld-supercard

### DIFF
--- a/src/lib/y2firewall/widgets/zone.rb
+++ b/src/lib/y2firewall/widgets/zone.rb
@@ -61,11 +61,6 @@ module Y2Firewall
       def store
         @zone.name = value
       end
-
-      # Sets the focus into this widget
-      def focus
-        Yast::UI.SetFocus(Id(widget_id))
-      end
     end
 
     # short name of zone.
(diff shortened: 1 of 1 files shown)


package/yast2-firewall.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        4.1.7
+Version:        4.1.8
 Release:        0
```
teclator: It depends on this yast/yast-yast2#889, isn't it? In other case the attributes will be exported, so what about requiring also the version used in that PR when merged?.

src/autoyast-rnc/firewall.rnc
```diff
@@ -129,14 +129,10 @@ zones =
       zone_short? &
       zone_description? &
       zone_target? &
-      fwd_forward_ports? &
```
teclator: Remember also to modify the schema accordingly

https://github.com/yast/yast-schema/commit/7d3744e785db5965e9af00087e80e240a1929004

teclator:
It LGTM just some small comment / question, so will approve it once answered / solved.

PR #119, opened 2019-01-21
AY Config: Fixed issues with importation and configuration changes 

teclator:
- [Trello Card](https://trello.com/c/5OWTqVV4/594-fate-324662-sle-15-sp1-yast-module-for-firewalld-supercard)
- [Fate#324662](https://fate.suse.com/324662)

This PR fixes these issues found by @jreidinger in AY config mode:

- [X] start autoyast config, import autoyast profile -> It prints firewall is not configured
- [X] start autoyast config, import autoyast profile, Edit -> it uses empty profile and not imported one for editing
- [X] start autoyast config, klone system, define custom zone and accept -> custom zone is not print by presenter ( not sure why )

### DIFF
--- a/test/lib/y2firewall/clients/auto_test.rb
+++ b/test/lib/y2firewall/clients/auto_test.rb
@@ -20,26 +20,25 @@
 # ------------------------------------------------------------------------------
 
 require_relative "../../../test_helper"
+require "y2firewall/dialogs/main"
 require "y2firewall/clients/auto"
 
 describe Y2Firewall::Clients::Auto do
   let(:firewalld) { Y2Firewall::Firewalld.instance }
+  let(:installed) { true }
   let(:autoyast) { double("Y2Firewall::Autoyast", import: true, export: {}) }
 
   before do
-    allow_any_instance_of(Y2Firewall::Firewalld::Api).to receive(:running?).and_return(false)
     subject.class.imported = false
+
     allow(firewalld).to receive(:read)
-    allow(firewalld).to receive(:installed?).and_return(true)
+    allow(firewalld).to receive(:installed?).and_return(installed)
     allow(subject).to receive(:autoyast).and_return(autoyast)
+    allow_any_instance_of(Y2Firewall::Firewalld::Api).to receive(:running?).and_return(false)
   end
 
   describe "#summary" do
     let(:installed) { false }
-
-    before do
-      allow(firewalld).to receive(:installed?).and_return(installed)
-    end
 
     context "when firewalld is not installed" do
       it "reports when firewalld is not available" do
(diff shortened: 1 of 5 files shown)


test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -229,6 +228,82 @@
...
+    let(:main_dialog) { instance_double("Dialog::Main", run: result) }
+
+    before do
+      allow(subject).to receive(:import)
+      allow(subject).to receive(:export).and_return(exported_profile)
+      allow(Y2Firewall::Dialogs::Main).to receive(:new).and_return(main_dialog)
+      allow(firewalld).to receive(:read?).and_return(already_read)
+    end
+
+    context "when the configuration is accepted" do
```
teclator: Mmm it says when it is accepted but the results seems to be the rejected ones

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -229,6 +228,82 @@
...
+
+        it "does not set autoyast config" do
+          expect(subject.class).to_not receive(:ay_config=)
+
+          subject.change
+        end
+      end
+    end
+
+    context "when the configuration is rejected" do
```
teclator: See above

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -229,6 +228,82 @@
...
+        let(:already_read) { true }
+
+        it "does not read it again" do
+          expect(firewalld).to_not receive(:read)
+
+          subject.change
+        end
+      end
+
+      context "but its configuation is not read" do
```
teclator:
```suggestion
      context "but its configuration has not been read" do
```

test/lib/y2firewall/clients/auto_test.rb
dgdavid: Oops! Wrong sentence. It seems that, accidentally, I switched them.

Thanks!

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -229,6 +228,82 @@
...
+        it "reads the minimal configuration" do
+          expect(firewalld).to receive(:read).with(minimal: true)
+
+          subject.change
+        end
+
+        it "keeps the previous configuration" do
```
teclator: The scenario that we want to cover is when a configuration has been imported previously and then we modify it. We want to ensure that the configuration is not lost.

So would test it importing a minimal profile. For example:

```xml
<zones config:type="list">
  <zone>
    <name>example</name>
    <services config:type="list">
      <service>http</service>
      <service>https</service>
      <service>ssh</service>
    </services>
  </zone>
</zones>  
```

teclator:
@dgdavid the test LGTM, just some comments. Thnx for taking care!!

test/lib/y2firewall/clients/auto_test.rb
dgdavid: Ok, let me improve/fix it :)

Thank you!

teclator: Although from other POV it is more an integration test than a unit test, so if it is too complicate just let it as it is.

dgdavid: Ok, take a look to https://github.com/yast/yast-firewall/pull/119/commits/19a1375f672a4be31d936689c83f0c87b0a6f4ae Does it look good for you? If so, I'll squash it in the previous commit. If not, I'll discard it and 

> just let it as it is.

Thanks!

PR #120, opened 2019-03-27
Export zone's name explicitly as it is not a common attribute (bsc#1130354)

teclator:
- It depends on https://github.com/yast/yast-yast2/pull/914
- https://bugzilla.suse.com/show_bug.cgi?id=1130354

### DIFF
--- a/src/lib/y2firewall/widgets/pages/zones.rb
+++ b/src/lib/y2firewall/widgets/pages/zones.rb
@@ -79,7 +79,6 @@ module Y2Firewall
             result = Dialogs::Zone.run(zone, new_zone:       true,
                                              existing_names: firewall.zones.map(&:name))
             if result == :ok
-              zone.relations.map { |r| zone.send("#{r}=", []) }
               firewall.zones << zone
               UIState.instance.select_row(zone.name)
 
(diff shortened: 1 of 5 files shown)


src/lib/y2firewall/widgets/pages/zones.rb
```diff
@@ -79,7 +79,6 @@ def handle
             result = Dialogs::Zone.run(zone, new_zone:       true,
                                              existing_names: firewall.zones.map(&:name))
             if result == :ok
-              zone.relations.map { |r| zone.send("#{r}=", []) }
```
jreidinger: what is goal of this change? commit message says what you did, but not why.

teclator: You are right, I should mention in the changelog or somewhere. 

So this initialization was probably needed before added it to the constructor, which was later done (later taking in account the date it was added here).

https://github.com/yast/yast-firewall/commit/873bbf5249d1cbeb13b9db981da948cf605f6689#diff-957421c569fb30a22c554130ec95497eR72

https://github.com/yast/yast-yast2/pull/840/files#diff-5f18f6aba33e047bd40791544760f288R65

jreidinger: ok, that makes sense.

PR #123, opened 2019-04-26
Cpu mitigations

jreidinger:


src/lib/y2firewall/clients/proposal.rb
```diff
@@ -106,15 +108,54 @@ def write
...
+
+        res = _("CPU Mitigations: ") + format("<a href=\"%s\">", LINK_CPU_MITIGATIONS) +
+          mitigations.to_human_string + "</a>"
+        log.info "mitigations output #{res.inspect}"
+        res
+      end
+
+      def bootloader_dialog
+        require "bootloader/config_dialog"
+        Yast.import "Bootloader"
```
lslezak: NP: As there is no dependency on `yast2-bootloader` (as this is only for the installer where we can assume it's always present) in might theoretically raise an exception in the `require` here. In that case the `ensure Yast::Wizard.CloseDialog` call would be wrong, the dialog would not be created yet.

I'd add a `begin` statement below.

src/lib/y2firewall/clients/proposal.rb
```diff
@@ -106,15 +108,54 @@ def write
...
+      # Returns nil if this part should be skipped
+      # @return [String] proposal html text
+      def cpu_mitigations_proposal
+        require "bootloader/bootloader_factory"
+        bl = ::Bootloader::BootloaderFactory.current
+        return nil if bl.name == "none"
+
+        mitigations = bl.cpu_mitigations
+
+        res = _("CPU Mitigations: ") + format("<a href=\"%s\">", LINK_CPU_MITIGATIONS) +
```
lslezak: NP: I'd prefer having also the closing tag in the `format` call:

```ruby
res = _("CPU Mitigations: ") + format("<a href=\"%s\">%s</a>", LINK_CPU_MITIGATIONS, 
  mitigations.to_human_string)
```

and we should probably `html_escape` the values, just to be safe (maybe escaping just the `mitigations.to_human_string` value would be enough):

```ruby
require "erb"

res = _("CPU Mitigations: ") + format("<a href=\"%s\">%s</a>",
  ERB::Util.html_escape(LINK_CPU_MITIGATIONS), 
  ERB::Util.html_escape(mitigations.to_human_string))
```

test/lib/y2firewall/clients/proposal_test.rb
```diff
@@ -24,14 +24,19 @@
...
+    # skip bootloader proposal to avoid build dependency on it
+    allow(subject).to receive(:cpu_mitigations_proposal)
+  end
+
   describe "#initialize" do
     it "instantiates a new proposal settings" do
       expect(Y2Firewall::ProposalSettings).to receive(:instance)
 
-      client
+      described_class.new
```
lslezak: Would be just calling `subject` here enough?

src/lib/y2firewall/clients/proposal.rb
jreidinger: ok, make sense.

src/lib/y2firewall/clients/proposal.rb
jreidinger: In fact here is format useless and can be directly included. I intentionally have it outside of translation. html escape won't work properly for href argument, for human string it makes some sense. I will adapt.

test/lib/y2firewall/clients/proposal_test.rb
jreidinger: no, as subject is already initialized in before, so expect never happen. I will add comment.

lslezak: Ah, I see...

PR #124, opened 2019-04-29
Sle15 sp1 merge

jreidinger:


package/yast2-firewall.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        4.1.11
+Version:        4.2.0
```
lslezak: This is the very same change as in the SP1 branch, right? Then we can keep the SP1 version, it should be changed only when doing TW or SP2 change not present in SP1...

jreidinger: but I modify spec, so spec is not same and I think in that case if htere is differ, we need to have different version, not?

package/yast2-firewall.spec
```diff
@@ -38,6 +38,8 @@ Requires:       yast2 >= 4.1.67
 
 # ButtonBox widget
 Conflicts:	yast2-ycp-ui-bindings < 2.17.3
+# CpiMitigations
+Conflicts:	yast2-bootloader < 4.2.1
```
jreidinger: this is different to SP1 version as in TW bootloader version is higher.

PR #127, opened 2020-02-26
Mode autoyast

teclator:
## Problem

In case of an autoinstallation with **VNC**, the server is not reachable during the second stage.

- https://trello.com/c/thlto0ec/1610-5-sles15-sp2-p2-1161319-build-1261-vnc-is-not-reachable-for-second-stage-during-autoyast-installation-on-zkvm

## Solution

Bring back the firewall installation finish client which opens the **VNC** or **SSH** server ports if firewalld is installed and the installation is done remotely through **VNC** or **SSH**.

## Test

- Tested manually

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Wed Feb 26 20:23:04 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
 
-- Autoyast: Added back the installation finish client for opening
-  the vnc or ssh port during the first stage when used for remote
+- AutoYaST: Added back the installation finish client for opening
+  the VNC or SSH port during the first stage when used for remote
   installations (bsc#1161319)
 - 4.2.3
 
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Feb 26 20:23:04 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Autoyast: Added back the installation finish client for opening
+  the vnc or ssh port during the first stage when used for remote
+  installations (bsc#1161319)
+- 4.2.3
+
+-------------------------------------------------------------------
 Tue Aug 27 18:00:06 CEST 2019 - schubi@suse.de
 
 - Set X-SuSE-YaST-AutoInstResource in desktop file (bsc#144894).
--- a/src/lib/y2firewall/clients/installation_finish.rb
+++ b/src/lib/y2firewall/clients/installation_finish.rb
@@ -24,6 +24,8 @@ require "y2firewall/firewalld"
 require "y2firewall/proposal_settings"
 require "installation/finish_client"
 
+Yast.import "Mode"
+
 module Y2Firewall
   module Clients
     # This is a step of base installation finish and it is responsible of write
@@ -49,7 +51,7 @@ module Y2Firewall
       end
 
       def modes
-        [:installation]
+        [:installation, :autoinst]
       end
 
       def write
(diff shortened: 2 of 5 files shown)


package/yast2-firewall.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Feb 26 20:23:04 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Autoyast: Added back the installation finish client for opening
```
imobachgs: Please, fix capitalization (AutoYaST, VNC, and SSH).

src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -63,14 +65,32 @@ def write
       # Modifies the configuration of the firewall according to the current
       # settings
       def configure_firewall
+        configure_service
```
imobachgs: I would rename the method to `configure_firewall_service`.

test/lib/y2firewall/clients/installation_finish_test.rb
```diff
@@ -69,26 +69,30 @@
...
       allow(proposal_settings).to receive("open_ssh").and_return(false)
     end
 
-    it "enables the firewalld service if enabled in the proposal" do
-      allow(proposal_settings).to receive("enable_firewall").and_return(true)
-      expect(firewalld).to receive("enable!")
+    context "during an installation" do
+      it "enables the firewalld service if enabled in the proposal" do
+        allow(proposal_settings).to receive("enable_firewall").and_return(true)
+        expect(firewalld).to receive("enable!")
```
imobachgs: np: Using a string looks strange. What about simply using a symbol (`:enable!`)? It is what we usually do.

test/lib/y2firewall/clients/installation_finish_test.rb
```diff
@@ -69,26 +69,30 @@
...
 
-      subject.send(:configure_firewall)
-    end
+        subject.send(:configure_firewall)
+      end
 
-    it "disables the firewalld service if disabled in the proposal" do
-      expect(firewalld).to receive("disable!")
+      it "disables the firewalld service if disabled in the proposal" do
+        expect(firewalld).to receive("disable!")
```
imobachgs: np: same than above

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: Yep, that is true. Will modify it.

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: Same than above xD

test/lib/y2firewall/clients/auto_test.rb
```diff
@@ -399,4 +399,10 @@
       end
     end
   end
+
+  describe "#packages" do
```
teclator: It was not related to the fix, but decreasing the test coverage is not nice :laughing: . So, this was added just for increasing the test coverage slightly.

test/lib/y2firewall/clients/installation_finish_test.rb
imobachgs: You have not modified it, right? :)

test/lib/y2firewall/clients/installation_finish_test.rb
imobachgs: Same than above :)

test/lib/y2firewall/clients/installation_finish_test.rb
teclator: Now, it is definitely modified :stuck_out_tongue_closed_eyes:

PR #133, opened 2020-07-20
Run autoyast configuration in the first stage

mchf:
## Problem ##

We want to run firewall configuration in the fist stage of installer. However, there is problem when installing over ssh / vnc - we don't want to shoot into own leg in such case.

## Solution ##

Uses https://github.com/yast/yast-autoinstallation/pull/653 which loads the firewall client in the first stage. The configuration is completely done during the first stage with one exception. When running over ssh / vnc enabling firewall is postponed to the second stage

## Tests ##

Manual, beside others:
* configure firewall rules & firewall is enabled after installation
* configure firewall rules & firewall is disabled after installation
* configure firewall rules & firewall is enabled after installation, installation over ssh
* configure firewall rules & firewall is disabled after installation, installation over ssh

TBD:
upload testsuite

### DIFF
--- a/src/lib/y2firewall/clients/auto.rb
+++ b/src/lib/y2firewall/clients/auto.rb
@@ -82,7 +82,6 @@ module Y2Firewall
       # @return [Boolean]
       def import(profile, merge = !Yast::Mode.config)
         self.class.profile = profile
-        # in fact, merge is currently not used elsewhere
         return false if merge && !read(force: false)
 
         if Yast::Stage.cont
@@ -176,6 +175,7 @@ module Y2Firewall
       # stage. Exceptions are installation over network (ssh / vnc) with second stage
       # enabled which requires more careful approach
       def import_first_stage(merge = !Yast::Mode.config)
+        log.info("Firewall: importing part of the profile which can be processed in first stage")
         # Obtains the default from the control file (settings) if not present.
         start_firewall_on_target if !need_second_stage_run?
 
@@ -189,6 +189,7 @@ module Y2Firewall
       #
       # Intended for finishing setup of exceptional cases. @see import_first_stage
       def import_second_stage
+        log.info("Firewall: second stage cleanup")
         true
       end
 
--- a/src/lib/y2firewall/clients/auto.rb
+++ b/src/lib/y2firewall/clients/auto.rb
@@ -203,15 +203,15 @@ module Y2Firewall
         # 3) firewall was configured (somehow) and started via AY profile
         remote_installer = Linuxrc.usessh || Linuxrc.vnc
         second_stage_required = self.class.profile.fetch("second_stage", false)
-        firewall_started = self.class.profile.fetch("enable_firewall", settings.enable_firewall)
+        firewall_started = firewall_profile.fetch("enable_firewall", settings.enable_firewall)
 
         remote_installer && second_stage_required && firewall_started
       end
 
       # Configures firewall service to run on target according to AY profile and product defaults
       def start_firewall_on_target
-        enable if self.class.profile.fetch("enable_firewall", settings.enable_firewall)
-        start if self.class.profile.fetch("start_firewall", false)
+        enable if firewall_profile.fetch("enable_firewall", settings.enable_firewall)
+        start if firewall_profile.fetch("start_firewall", false)
       end
 
       # Read the minimal configuration from firewalld, w/o dropping available configuration
@@ -296,6 +296,12 @@ module Y2Firewall
       # @return [Y2Firewall::ProposalSettings]
       def settings
         ProposalSettings.instance
+      end
+
+      # @return [Hash<String, any>] firewall section of the profile
+      def firewall_profile
+        return {} if !self.class.profile
+        self.class.profile.fetch("firewall", {})
       end
 
       # Set that the firewall has to be enabled when writing
--- a/src/lib/y2firewall/clients/auto.rb
+++ b/src/lib/y2firewall/clients/auto.rb
@@ -91,7 +91,7 @@ module Y2Firewall
           # stage is enabled and installation runs over network (ssh / vnc)
           import_second_stage
         else
-          import_first_stage(profile, merge)
+          import_first_stage(merge)
         end
       end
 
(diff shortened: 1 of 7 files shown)


src/lib/y2firewall/clients/auto.rb
```diff
@@ -79,16 +81,18 @@ def summary
...
         self.class.profile = profile
         return false if merge && !read(force: false)
 
-        # Obtains the default from the control file (settings) if not present.
-        enable if profile.fetch("enable_firewall", settings.enable_firewall)
-        start if profile.fetch("start_firewall", false)
-        autoyast.import(profile)
-        check_profile_for_errors
-        imported
+        if Yast::Stage.cont
```
teclator, line 88: I would move this responsibility into the autoinstallation module. In fact, one of the goals was to remove the section once imported.

https://github.com/yast/yast-autoinstallation/pull/653#discussion_r457214395

At that point we know exactly if it should be configured during the first stage or not. One of the main problems is that we need the package already installed to read the default configuration and merge with the one from the profile.

src/lib/y2firewall/clients/auto.rb
```diff
@@ -215,9 +270,20 @@ def check_profile_for_errors
 
       # Depending on the profile it activates or deactivates the firewalld
       # service
+      #
+      # Note: it always activates the service on the target
       def activate_service
-        enable? ? firewalld.enable! : firewalld.disable!
```
teclator, line 219: Dos not work by some reason?

src/lib/y2firewall/clients/auto.rb
```diff
@@ -79,16 +81,18 @@ def summary
       #
       # @param profile [Hash] firewall profile section to be imported
       # @return [Boolean]
-      def import(profile, merge = !Yast::Mode.config)
+      def import(profile, merge = !Yast::Mode.config || !Yast::Mode.auto)
```
teclator, line 84: In autoinstallation the merge was also used as by default each zone contains some predefined services / ports open.

src/lib/y2firewall/clients/auto.rb
mchf: it was and is my big dilemma. Putting it directly into AY is not problem. But on the other hand I have a feeling that it is also responsibility of the firewall module. And I'm still unsure ;-)

I'll prepare also a version with logic moved to AY, possibly it will add some points and break 50 : 50 problem ;-)

src/lib/y2firewall/clients/auto.rb
mchf: I don't remmember. I think there was a problem but I'm not sure now. Will retest.

src/lib/y2firewall/clients/auto.rb
mchf: But since the import is called by AY at the beginning of the first stage now, it is called before the configuration exists (firewalld is not present at this point). It is one thing I wanted to open a discussion about here ;-)

I also have a version where even import is called from the finish client and there it works just fine.

So as I currently see following options
1) keep it as it is here, but there can be issues with non merging configurations
2) go back to second approach mentioned above, but it was using ```Profile.current```
3) call firewall's import from AY but from different place than others - somewhere around calling finish clients
4) keep things as they are now - keep second stage for firewall mandatory in all cases
5) I missed something

teclator: > But since the import is called by AY at the beginning of the first stage now, it is called before the configuration exists (firewalld is not present at this point). It is one thing I wanted to open a discussion about here ;-)

If I'm not wrong that was handled by the client. If the package was not installed then it was not imported and tried to import later and that is why the profile was stored as a class variable.

> 
> I also have a version where even import is called from the finish client and there it works just fine.

Yep, that should be the behavior.

> 
> So as I currently see following options
> 
> 1. keep it as it is here, but there can be issues with non merging configurations

Should be merged during the finish write.

> 2. go back to second approach mentioned above, but it was using `Profile.current`
> 3. call firewall's import from AY but from different place than others - somewhere around calling finish clients
> 4. keep things as they are now - keep second stage for firewall mandatory in all cases
> 5. I missed something

src/lib/y2firewall/clients/auto.rb
```diff
@@ -166,6 +170,57 @@ def modified?
 
     private
 
+      # Import method for the first stage.
+      #
+      # Except a few edge cases complete firewall configuration should be done in first
+      # stage. Exceptions are installation over network (ssh / vnc) with second stage
```
teclator, line 176: I would do always the import and configuration during the first stage but opening the ports as we do during the installation only if the second_stage is enabled.

At the end of the second stage we should close opened ports that were not declared in the profile, and this should be probably done checking some mark as a temporal file.

src/lib/y2firewall/clients/auto.rb
mchf: > If I'm not wrong that was handled by the client. If the package was not installed then it was not imported and tried to import later and that is why the profile was stored as a class variable.

then there was a bug. In short in profile where I wanted firewall to be enabled after installation, the import ended with false (because firewalld was not installed) and later write didn't change anything. I know there is ```import_if_needed``` method, however the firewall simply was not enabled after installation. So, I'm taking note to investigate that.
> 
> > I also have a version where even import is called from the finish client and there it works just fine.
> 
> Yep, that should be the behavior.

i know but worked only if I placed the import into the client explicitly. See above
> 
> > So as I currently see following options
> > 
> > 1. keep it as it is here, but there can be issues with non merging configurations
> 
> Should be merged during the finish write.

ok, as described above, I'll check what is wrong that this didn't work.
>

src/lib/y2firewall/clients/auto.rb
mchf: Well I didn't want to fiddle with logic for adding / removing rules for ssh / vnc so I simply postpone enabling firewall into second stage if needed. Worked fine and logic is simple. It makes the system open in some cases during the second stage but it should be "short" period and in fact behaves the same like before in all cases.

mchf:
Replaced by PR#135

PR #135, opened 2020-07-28
Run autoyast configuration in the first stage

mchf:
Reworked previous version, see https://github.com/yast/yast-firewall/pull/133 for details

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Aug 13 07:52:47 UTC 2020 - Michal Filka <mfilka@suse.com>
 
-- AutoYaST: moved the configuration into first stage
+- AutoYaST: moved the configuration into first stage (bsc#1173624)
 - 4.3.4
 
 -------------------------------------------------------------------
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Aug 13 07:52:47 UTC 2020 - Michal Filka <mfilka@suse.com>
+
+- AutoYaST: moved the configuration into first stage
+- 4.3.4
+
+-------------------------------------------------------------------
 Tue Aug 11 12:05:15 CEST 2020 - schubi@suse.de
 
 - AutoYaST: Added supplements: autoyast(firewall) into the spec file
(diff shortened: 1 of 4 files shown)


src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -55,6 +56,13 @@ def modes
       end
 
       def write
+        Yast.import "Mode"
+
+        if Yast::Mode.auto
```
teclator, line 61: Quite simpler than previous PR but.. how it works when there is no firewall section at all and the installation is done through ssh or vnc? I guess it will block the traffic, see https://github.com/yast/yast-firewall/pull/127 and why we needed to add the auto-installation mode back for the finish client even before moving it to the first stage.

Thus, if no section is added at all I would expect the proposal settings to be used instead.

mchf: hm, good point. didn't try. Will check and adapt.

mchf: so, in fact it worked (for ssh), however involving product settings was a good point, so I've adapted the path

package/yast2-firewall.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Jul 28 11:01:02 UTC 2020 - Michal Filka <mfilka@suse.com>
+
+- AutoYaST: moved the configuration into first stage
```
teclator, line 4: Miss bug number entry, I created a bug to be used by all the modules which moved the configuration to the first stage

teclator, line 4:
```suggestion
- AutoYaST: moved the configuration into first stage (bsc#1173624)
```

src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -55,6 +57,16 @@ def modes
       end
 
       def write
+        Yast.import "Mode"
```
teclator, line 60: It is already imported...

mchf: right

test/lib/y2firewall/clients/installation_finish_test.rb
```diff
@@ -62,6 +62,33 @@
         expect(subject.write).to eq true
       end
     end
+
+    context "when running in AutoYaST" do
```
teclator, line 66: Wanted to ask about updating the unit test but you already did ... so, thnx ;)

mchf: welcome ... I can read you as a book ;-)

PR #136, opened 2020-08-11
AutoYaST: Added supplements: autoyast(firewall) into the spec file in…

schubi2:
… order to install this packages if the section has been defined in the AY configuration file (bsc#1146494).


lslezak:
Actually the firewall module is being moved to the first stage, see #135. If we move it then we will not need this one, right?

schubi2:
We have decided to let it :-)

PR #137, opened 2020-10-09
Prevent the user to disable or close ssh access when only SSH key based authentication is used.

teclator:
## Problem

When the user selects to authenticate using only SSH key based authentication we propose to enable and open the SSH ports (since https://github.com/yast/yast-firewall/pull/103)

If the user change the SSH proposal, then the user might not be allowed to login.

- https://trello.com/c/Mr566lSn/2059-3-shouldhave-add-ssh-enable-disable-to-root-password-dialog

## Solution

Warn the user about the possible access problem

SSH_warning


package/yast2-firewall.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Oct  9 10:42:32 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Warn the user when SSH open port is closed or the the service is
+  disable and the configured authentication is only based on SSH
```
imobachgs, line 5: disabled

imobachgs (approved):
Just a typo. Otherwise, LGTM.

package/yast2-firewall.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Oct  9 10:42:32 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Warn the user when SSH open port is closed or the the service is
```
lslezak, line 4:
```suggestion
- Warn the user when the SSH port is closed or the service is
```

package/yast2-firewall.changes
teclator, line 5:
```suggestion
  disabled and the configured authentication is only based on SSH
```

PR #139, opened 2020-10-16
Fix enabling firewall

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1177778

### DIFF
--- a/src/lib/y2firewall/clients/installation_finish.rb
+++ b/src/lib/y2firewall/clients/installation_finish.rb
@@ -73,6 +73,9 @@ module Y2Firewall
       # Convenience method to enable / disable the firewalld service depending
       # on the proposal settings
       def configure_firewall_service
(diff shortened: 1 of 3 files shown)


src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -73,6 +73,9 @@ def configure_firewall
       # Convenience method to enable / disable the firewalld service depending
       # on the proposal settings
       def configure_firewall_service
+        # do not run in autoyast as it is done in second stage (bsc#1177778)
+        return if Yast::Mode.auto
```
teclator, line 77: Would be nice to configure according to the profile in case of defined section but that is or should be already done in SP3 as it is already configured during the first stage. Thus, for the time being it is ok and nice for SP2.

teclator (approved):
LGTM

PR #140, opened 2020-10-19
Sp2 merge

jreidinger:
merge of https://github.com/yast/yast-firewall/pull/139

regarding moving firewall config to first stage, I would let it to own PBI.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,5 @@
 require "yast/rake"
 
-Yast::Tasks.submit_to :sle15sp2
-
 Yast::Tasks.configuration do |conf|
   # lets ignore license check for now
   conf.skip_license_check << /.*/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,3 @@
-FROM registry.opensuse.org/yast/sle-15/sp2/containers/yast-ruby
+FROM registry.opensuse.org/yast/head/containers/yast-ruby:latest
 COPY . /usr/src/app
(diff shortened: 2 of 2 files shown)


Rakefile
```diff
@@ -1,5 +1,7 @@
 require "yast/rake"
 
+Yast::Tasks.submit_to :sle15sp2
```
teclator, line 3: this should be removed or revoked the same as the Dockerfile

src/lib/y2firewall/clients/installation_finish.rb
```diff
@@ -81,6 +81,9 @@ def configure_firewall
       # Convenience method to enable / disable the firewalld service depending
       # on the proposal settings
       def configure_firewall_service
+        # do not run in autoyast as it is done in second stage (bsc#1177778)
+        return if Yast::Mode.auto
```
teclator, line 85: As commented in #139 it probably should not be done in master as it should be already handled.

teclator: https://github.com/yast/yast-firewall/blob/a4a998c6e6e215e6f6d0492ae644e70b7567e870/src/lib/y2firewall/clients/installation_finish.rb#L63

teclator: After checked it is ok to be added.

teclator (approved):
LGTM

PR #144, opened 2021-02-03
Selinux proposal

jreidinger:
trello: https://trello.com/c/g94xQBDM/2159-8-musthave-selinux-and-security-refac

related: https://github.com/yast/yast-security/pull/83

### DIFF
--- a/package/yast2-firewall.changes
+++ b/package/yast2-firewall.changes
@@ -1,8 +1,15 @@
 -------------------------------------------------------------------
+Thu Feb  4 09:14:13 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
+- Add to firewall/security proposal option to setup selinux if
+  given product require it. (jsc#SLE-17427)
+- 4.2.6
+
+-------------------------------------------------------------------
 Fri Oct 16 15:15:49 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
 
 - Do not enable firewall during first stage of AutoYaST
-  (bsc#1177778) 
+  (bsc#1177778)
 - 4.2.5
 
 -------------------------------------------------------------------
--- a/package/yast2-firewall.spec
+++ b/package/yast2-firewall.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-firewall
-Version:        4.2.5
+Version:        4.2.6
 Release:        0
 Summary:        YaST2 - Firewall Configuration
 Group:          System/YaST
@@ -26,8 +26,10 @@ Url:            https://github.com/yast/yast-firewall
 
 Source0:        %{name}-%{version}.tar.bz2
 
-BuildRequires:  perl-XML-Writer update-desktop-files yast2-testsuite
+BuildRequires:  update-desktop-files
 BuildRequires:  yast2-devtools >= 4.2.2
+# for proposing selinux
+BuildRequires:  yast2-security >= 4.2.15
 # Removed zone name from common attributes definition
 BuildRequires:  yast2 >= 4.1.67
 BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
--- a/package/yast2-firewall.spec
+++ b/package/yast2-firewall.spec
@@ -29,7 +29,7 @@ Source0:        %{name}-%{version}.tar.bz2
 BuildRequires:  update-desktop-files
 BuildRequires:  yast2-devtools >= 4.2.2
 # for proposing selinux
-BuildRequires:  yast2-security >= 4.2.16
+BuildRequires:  yast2-security >= 4.2.17
 # Removed zone name from common attributes definition
 BuildRequires:  yast2 >= 4.1.67
 BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
--- a/src/lib/y2firewall/dialogs/proposal.rb
+++ b/src/lib/y2firewall/dialogs/proposal.rb
@@ -41,34 +41,8 @@ module Y2Firewall
       end
 
       def contents
-        res = VBox(
-          Frame(
-            _("Firewall and SSH service"),
-            HSquash(
-              MarginBox(
-                0.5,
-                0.5,
-                VBox(
-                  Widgets::FirewallSSHProposal.new(@settings)
-                )
-              )
-            )
-          )
-        )
-
-        if @settings.selinux_config.configurable?
-          res.params << Frame(
-            _("SELinux"),
-            MarginBox(
-              0.5,
-              0.5,
-              VBox(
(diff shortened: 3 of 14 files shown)


src/lib/y2firewall/proposal_settings.rb
```diff
@@ -59,6 +59,7 @@ def initialize
       # chose a different one in the proposal
       @default_zone = "public"
       @selinux_config = Y2Security::SelinuxConfig.new
+      @selinux_config.save # lets write the proposal to sync the initial state
```
dgdavid, line 62: But current implementation of [SelinuxConfig#save](https://github.com/yast/yast-security/blob/2ccd5c23b7bc1f6fd77d6b9e8ff55d199a1b6258/src/lib/y2security/selinux_config.rb#L128-L148) will not write the initial state unless [SelinuxConfig#configurable?](https://github.com/yast/yast-security/blob/2ccd5c23b7bc1f6fd77d6b9e8ff55d199a1b6258/src/lib/y2security/selinux_config.rb#L150-L159)

Is that ok?

jreidinger: yes, it is exactly what I need. As if it is not configurable, we do not touch it, so it is basically nothing and we do not display anything, neither write.

package/yast2-firewall.spec
```diff
@@ -26,8 +26,10 @@ Url:            https://github.com/yast/yast-firewall
 
 Source0:        %{name}-%{version}.tar.bz2
 
-BuildRequires:  perl-XML-Writer update-desktop-files yast2-testsuite
+BuildRequires:  update-desktop-files
 BuildRequires:  yast2-devtools >= 4.2.2
+# for proposing selinux
+BuildRequires:  yast2-security >= 4.2.15
```
dgdavid, line 32: So, we should merge https://github.com/yast/yast-security/pull/83, right?

src/lib/y2firewall/dialogs/proposal.rb
```diff
@@ -55,6 +55,21 @@ def contents
             )
           )
         )
+
+        if @settings.selinux_config.configurable?
+          res.params << Frame(
```
dgdavid, line 60: Curious. 

NP: I would have used a `#selinux_content` method returning `Empty()` when not configurable and called it inside the above `VBox`. Much readable to me, but maybe it does not work.

dgdavid (approved):
LGTM, just added some comments but it looks good anyway. Thanks!

package/yast2-firewall.spec
jreidinger: yes, it is dependency, otherwise test cannot be run

src/lib/y2firewall/dialogs/proposal.rb
jreidinger: I worry that Empty can break a bit layout as Empty is empty space on screen and no noop. That is why I dynamicaly add that frame.

jreidinger:
It is still failing as in OBS is not the new version of security. It probably needs to be submitted together there?

dgdavid:
> It is still failing as in OBS is not the new version of security. It probably needs to be submitted together there?

I submitted it with fixed dependencies two hours ago. Not sure when will be available in the container.

jreidinger:
@lslezak when playing with PackagesProposal for pattern I get this failure when pattern is not available:

selinux_failure2

I think that issue is caused by fact that statuses does not find anything as such resolvable is not found?

jreidinger:
OK, so for layout I tried HVCenter approach and Stretch one.

HVCenter does not look like I wanted:
selinux_hvcenter
and here is Stretch one ( final one )
selinux_stretches

package/yast2-firewall.changes
```diff
@@ -1,8 +1,15 @@
+-------------------------------------------------------------------
+Thu Feb  4 09:14:13 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
+- Add to firewall/security proposal option to setup selinux if
+  given product require it. (jsc#SLE-17427)
```
dgdavid, line 5: NP: I'm using jsc#SMO-20, and jsc#SLE-17342. The latest is the DEV task.

package/yast2-firewall.spec
```diff
@@ -26,8 +26,10 @@ Url:            https://github.com/yast/yast-firewall
 
 Source0:        %{name}-%{version}.tar.bz2
 
-BuildRequires:  perl-XML-Writer update-desktop-files yast2-testsuite
+BuildRequires:  update-desktop-files
 BuildRequires:  yast2-devtools >= 4.2.2
+# for proposing selinux
+BuildRequires:  yast2-security >= 4.2.16
```
dgdavid, line 32: 4.2.17 includes file exceptions handlers and returns no patterns when SELinux mode is disabled.

src/lib/y2firewall/widgets/proposal.rb
```diff
@@ -200,5 +200,30 @@ def help
...
+    # Widget to set SELinux mode
+    class SelinuxMode < CWM::ComboBox
+      def initialize(settings)
+        textdomain "firewall"
+
+        @settings = settings
+      end
+
+      def label
+        _("SELinux Mode")
```
dgdavid, line 213: As said, maybe in the context of the proposal  "SELinux" here is a bit redundant.

dgdavid (approved):
LGTM, just np/minor comments

src/lib/y2firewall/widgets/proposal.rb
jreidinger: yeah, we also discuss it with @kobliha my argument is that SELinux Mode is easily googable, on other hand selinux is already in frame. So looks like you see it also duplicite, so lets remove it.

dgdavid: > SELinux Mode is easily googable

Should we add some useful help to the dialog?

package/yast2-firewall.changes
jreidinger: too many jira entries. I am using epic mentioned in dev task

src/lib/y2firewall/widgets/proposal.rb
```diff
@@ -200,5 +200,43 @@ def help
...
+        self.value = @settings.selinux_config.mode.id
+      end
+
+      def store
+        @settings.selinux_config.mode = value
+      end
+
+      def help
+        _(
+          "<p>Sets default selinux mode. Modes are: <ul>" \
```
dgdavid, line 231:
```suggestion
          "<p>Sets default SELinux mode. Modes are: <ul>" \
```

dgdavid (approved):
Still looking good :) Just a minor suggestion that, of course, you can dismiss :)

Thanks a lot!

PR #150, opened 2021-06-17
Use Security Settings

joseivanlopez:
## Problem

AutoYaST client is still using *Y2Firewall::ProposalSettings*, but it should use *Installation::SecuritySettings*. Note that *Installation::SecuritySettings* was already adapted to use the new *Y2Users* API:

*https://github.com/yast/yast-installation/pull/969

## Solution

Adapt code to use *Installation::SecuritySettings*. Moreover, unused clients were removed: *firewall_finish* and *firewall_proposal*.

## Test

* Unit tests adapted
* Manually tested


src/lib/y2firewall/proposal_settings.rb
```diff
@@ -165,7 +164,18 @@ def wanted_open_vnc?
     #   key in order to log into the system. In such a case, we need to enable the SSH
     #   service (including opening the port).
     def only_public_key_auth
-      Yast::UsersSimple.GetRootPassword.empty?
+      password = root_user&.password_content || ""
```
jreidinger, line 167: I am not sure if it is still in use. Please check as firewall_proposal is dead and security_proposal is replacement now.

dgdavid: That's true. `firewall_proposal` was replaced by `security_proposal` few months ago. However, it was not completely removed. Shall we do it now?

For more context, see

* https://github.com/yast/skelcd-control-leanos/pull/74
* https://github.com/yast/skelcd-control-openSUSE/pull/228/commits/b4fe2ed1b6fc78638a5fefbe90ae69122d5c033d
* https://github.com/yast/skelcd-control-openSUSE/pull/232

jreidinger: yes, that would be great. I do not remove it to have smoother migration.

joseivanlopez: I did not know. Ok, I will check.

joseivanlopez: It seems that it is still used by AutoYaST:

* https://github.com/yast/yast-firewall/blob/master/src/lib/y2firewall/clients/auto.rb#L240
* https://documentation.suse.com/sles/15-SP3/single-html/SLES-autoyast/#CreateProfile-firewall-example

teclator: Removing the **firewall proposal client** and **firewall_finish client** looks reasonable if not used by third party clients, about the referenced **settings** it should use the security ones because otherwise it could ignore changes done in the confirmation dialog.

We probably should also clarify the comment at https://github.com/yast/yast-firewall/blob/master/src/lib/y2firewall/clients/auto.rb#L84 as it takes into account the option from the profile or the one from the settings which by default will be the control file one but could be modified manually by the user.

https://github.com/yast/yast-firewall/blob/master/src/lib/y2firewall/proposal_settings.rb#L59

https://github.com/yast/yast-installation/blob/master/src/lib/installation/security_settings.rb#L76
https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/security_proposal.rb#L126

dgdavid: > It seems that it is still used by AutoYaST:

Screenshot_openSUSE-Leap_2021-06-21_08:55:08

teclator: As we commented, the proposal_settings should be the security ones that probably were already adapted.

And when the firewall settings are imported the first time it should modify the proposal settings:

https://github.com/yast/yast-autoinstallation/blob/607def7568218a7855ca49151f6d0b6cb5f38487/src/lib/autoinstall/clients/inst_autosetup.rb#L384

But should respect them when writing the config, something that it is currently not doing If I'm not wrong.

https://github.com/yast/yast-installation/blob/10c899f646de091539f80002adebd8bb79df250c/src/lib/installation/clients/security_finish.rb#L177

jreidinger (approved):
+24 −1,325 is always good score for pr :+1:

teclator (approved):
LGTM

PR #151, opened 2021-06-21
Fix test

joseivanlopez:
## Problem

There is a failure requiring *installation/security_settings* after merging https://github.com/yast/yast-firewall/pull/150. But *yast2-installation* cannot be added as build dependency because *yast2-installation* already depends on *yast2-firewall*.

* https://ci.opensuse.org/job/yast-yast-firewall-master/47/console

## Solution

Mock `Installation::SecuritySettings` in order to avoid dependency of *yast2-installation*.

### DIFF
--- a/test/lib/y2firewall/clients/auto_test.rb
+++ b/test/lib/y2firewall/clients/auto_test.rb
@@ -1,23 +1,23 @@
 #!/usr/bin/env rspec
 
-# ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jreidinger (approved):

teclator (changes requested):
If I'm not wrong it is not enough and the auto client will fail because it cannot include **installation/security_settings**

teclator (approved):
LGTM

